PR #27, opened 2020-07-06
Find my safe

nicosampler:


src/apps/FindMySafe/index.tsx
```diff
@@ -0,0 +1,299 @@
...
+    try {
+      const res = await getBalances(safe, safeInfo!.network);
+      setBalances(res);
+      setLoadingBalances(false);
+    } catch (error) {
+      setLoadingBalances(false);
+    }
+  };
+
+  const getBalancesContent = () => {
```
mikheevm, line 198: this file is getting too big and hard to follow, is it possible to split it into separate components? for example, you can extract this to a separate component `Balances`

nicosampler: Sorry, wasn't ready to review yet. I created it to test something and I forget to put it on WIP

nicosampler: @mikheevm ready for review now!

rmeissner:
May I ask why this should be a safe app? (You never use the safe apps functionality here)

mikheevm:
I agree with @rmeissner, there's a lot of requests of users trying to find their safe address. This app, as I understand, aims to help, but in order to use it, one has to create a new safe, which is weird imo

nicosampler:
yes, guys, you are right, I missed that use case and seems to be the most important. I was thinking about users that already have a safe. Luckily this is pretty simple to move to the safe, let's ask @lukasschor his opinion.

nicosampler:
It works as a standalone app as well. You can check it here https://upbeat-kowalevski-f8a155.netlify.app/find-my-safe

lukasschor:
I agree with Mikhail & Richard. And we have a ticket for introducing a feature natively to automatically suggest Safes to the user where he/she is an owner of: https://github.com/gnosis/safe-react/issues/510

However, I still think it's good that we have this Safe App because of a few reasons:
- The nice thing about Safe Apps is that we don’t have to touch safe-react code
- Less overhead (incl. design, etc.) and easier to implement
- QA does not have to check if the feature potentially broke anything somewhere else
- It works as a standalone app that we can already refer to from the Help Center or support channels
- It also showcases that there are potential Safe Apps that don't necessarily need to be "Transaction crafter" apps

rmeissner:
For the showcase usecase I find the erc721 collection a better example. Also as @nicosampler a standalone website makes more sense for this.

Also I would like to know the purpose of this repo? Should it be for examples or for default apps developed by gnosis? There is a difference in maintenance.

nicosampler:
@rmeissner sure, could you create a repo?

rmeissner:
https://github.com/gnosis/safe-finder (we can rename it if required, but I would like to have something that starts with `safe` ;) )

mikheevm:
@rmeissner What do you think about adding it to the interface itself? Maybe on the load screen we can include a button "find my safe", or on the welcome screen. I don't think it's very hard

rmeissner:
I don'T really care, this is a decision that needs to be made by @lukasschor

lukasschor:
I would propose to put this on hold for now, and think about how to integrate this as a native feature instead.

rmeissner:
I would still close this PR in this case. @nicosampler feel free to move the code to the new repo ;)

Issue #29, opened 2020-07-14
Transaction builder can't load the ABI for Uniswap


joshfraser:
Since there isn't yet an official Uniswap app, I wanted to try to supply liquidity manually using the transaction builder. I got the following error message despite their ABI being [readily available](https://unpkg.com/@uniswap/v2-periphery@1.1.0-beta.0/build/IUniswapV2Router02.json) and [verified on Etherscan](https://etherscan.io/address/0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D#readContract):

<img width="1068" alt="Screen Shot 2020-07-13 at 8 42 32 PM" src="https://user-images.githubusercontent.com/982234/87381397-784fc680-c549-11ea-8441-6f4a2e03e907.png">

lukasschor:
Thanks. This seems to be a bug in the Transaction builder app.

If you don't necessarily require to batch transactions. You can also use the native "Contract Interaction" feature of the Safe Multisig to call the addLiquidity function. As there this bug does not exist. 

Here's more details on how to use this feature: https://help.gnosis-safe.io/en/articles/3738081-contract-interactions

lukasschor:
@nicosampler We need to update the transaction builder app to accommodate the fact that with Solidity version 0.6.0., the `payable` and `constant` fields have been deprecated. 

Some information on this:

https://github.com/ethereum/solidity/releases/tag/v0.6.0

https://solidity.readthedocs.io/en/v0.5.17/abi-spec.html#json vs. https://solidity.readthedocs.io/en/v0.6.11/abi-spec.html#json

cstrangedk:
@lukasschor having a similar problem with other ABIs with the TxBuilder... and the "Contract Interaction" feature also fails with the same ABIs.

lukasschor:
Any particular contracts that fail for you? Especially related to the "Contract Interaction" feature, as there we already update to be compatible with Solidity v0.6.0.

cstrangedk:
Just retested, apologies - contract interaction feature works fine for Solidity v0.6.0.  TxBuilder is only failing.

nicosampler:
@lukasschor this problem should be solved with this PR https://github.com/gnosis/safe-react-apps/pull/31.

PR #31, opened 2020-09-02
TX-Builder: fix filter methods condition

nicosampler:
closes #1258.
This should also fix https://github.com/gnosis/safe-react-apps/issues/29.

### DIFF
--- a/src/apps/TxBuilder/hooks/useServices/interfaceRepository.ts
+++ b/src/apps/TxBuilder/hooks/useServices/interfaceRepository.ts
@@ -53,7 +53,7 @@ class InterfaceRepository {
           return false;
         }
 
-        return !!e.constant === false;
+        return !e.constant;
       })
       .map((m: any) => {
         return { inputs: m.inputs, name: m.name, payable: m.payable || false };
--- a/src/apps/TxBuilder/hooks/useServices/interfaceRepository.ts
+++ b/src/apps/TxBuilder/hooks/useServices/interfaceRepository.ts
@@ -48,7 +48,13 @@ class InterfaceRepository {
     const abi = JSON.parse(abiString);
 
     const methods = abi
-      .filter((e: any) => e.constant === false)
+      .filter((e: any) => {
+        if (["pure", "view"].includes(e.stateMutability)) {
+          return false;
+        }
+
+        return !!e.constant === false;
+      })
       .map((m: any) => {
         return { inputs: m.inputs, name: m.name, payable: m.payable || false };
       });


src/apps/TxBuilder/hooks/useServices/interfaceRepository.ts
```diff
@@ -48,7 +48,13 @@ class InterfaceRepository {
     const abi = JSON.parse(abiString);
 
     const methods = abi
-      .filter((e: any) => e.constant === false)
+      .filter((e: any) => {
+        if (["pure", "view"].includes(e.stateMutability)) {
+          return false;
+        }
+
+        return !!e.constant === false;
```
rmeissner, line 56: If I understand this correctly this means if `e.constant` is `undefined` or `false` then we can use it, right? (my JS is not super good for this kind of magic :P)

nicosampler: `!!` is to handle falsy values, `constant` is optional (sometimes could come as undefined), in that case, we should assume it's false.

mikheevm, line 56: Was !e.constant not enough?

nicosampler: nop, we need to filter then the value **is** false

mikheevm: Sorry, what?

rmeissner: `e.constant` is of type `boolean | undefined` and we can use the function in the ui if `e.constant` is `false` or `undefined`. If it is `undefined` the `!!` will convert it to `false` and for `false` it will stay `false`. Therefore the `!!` is required.

nicosampler: this condition runs inside a filter, so, the current item needs to be included in the result when `e.constant` is false or undefined.

!!e.constant === false => true is different to !!e.constant if e.constant is false.

https://solidity.readthedocs.io/en/v0.5.3/abi-spec.html#json

rmeissner: I think what mikhail meant is that `!e.constant` actually evaluated to the same. if `e.constant` is `undefined` then `!e.constant` is `true` and if `e.constant` is `false` then `!e.constant` is also true.

rmeissner: acain this `undefined` stuff is dark js magic and I don't like it, so I am not making any promises that what I am writing is correct :P

mikheevm: That's exactly what I meant Richard, `!` will convert the value to its Boolean equivalent and perform a logical NOT operation

`!!` is used to convert a value to its Boolean equivalent, it does the same thing as `!` first, and then performs one more logical NOT operation on a value
https://link.medium.com/hSsK8ddDs9

Since we then compare the received value to false, `!` seems to be enough and IMO opinion it's better for understanding and readability (as simple as a developer has to follow one Boolean operation instead of 3)

mikheevm: Sorry Nico I had to leave this here
image

nicosampler: changed!

mikheevm (approved):
👍

PR #32, opened 2020-09-08
TX-Builder: Add support for Array and Tuple Values

nicosampler:
* Added Support for Array and Tuples values.
* Added a helper with values examples.
* Improve error messages.

image

In order to test it, I created this [contract](0x23b7f26872c2168814dc9d955bdade2961ea17f6) in Rinkeby. 
This is the ABI:
```
[
	{
		"inputs": [],
		"name": "counter",
		"outputs": [
			{
				"internalType": "uint8",
				"name": "",
				"type": "uint8"
			}
		],
		"stateMutability": "view",
		"type": "function"
	},
	{
		"inputs": [
			{
				"internalType": "uint256",
				"name": "value",
				"type": "uint256"
			}
		],
		"name": "swap",
		"outputs": [],
		"stateMutability": "nonpayable",
		"type": "function"
	},
	{
		"inputs": [
			{
				"internalType": "address",
				"name": "value",
				"type": "address"
			}
		],
		"name": "swap2",
		"outputs": [],
		"stateMutability": "nonpayable",
		"type": "function"
	},
	{
		"inputs": [
			{
				"internalType": "uint256",
				"name": "value",
				"type": "uint256"
			}
		],
		"name": "swap3",
		"outputs": [],
		"stateMutability": "nonpayable",
		"type": "function"
	},
	{
		"inputs": [
			{
				"internalType": "uint256[]",
				"name": "value",
				"type": "uint256[]"
			}
		],
		"name": "swap4",
		"outputs": [],
		"stateMutability": "nonpayable",
		"type": "function"
	},
	{
		"inputs": [
			{
				"components": [
					{
						"internalType": "uint256",
						"name": "amount",
						"type": "uint256"
					},
					{
						"internalType": "string",
						"name": "name",
						"type": "string"
					}
				],
				"internalType": "struct YourContract.testStruct[]",
				"name": "values",
				"type": "tuple[]"
			}
		],
		"name": "swap5",
		"outputs": [],
		"stateMutability": "nonpayable",
		"type": "function"
	}
]
```


src/apps/TxBuilder/components/Dashboard.tsx
```diff
@@ -151,7 +155,9 @@ const Dashboard = () => {
       description = method.name + " (";
       for (let i = 0; i < method.inputs.length; i++) {
         const cleanValue = inputCache[i] || "";
-        cleanInputs[i] = cleanValue;
+        cleanInputs[i] =
+          // eslint-disable-next-line
+          cleanValue.charAt(0) === "[" ? eval(cleanValue) : cleanValue;
```
fernandomg, line 160: Is there a chance to avoid using `eval`? If it's not, maybe adding a comment explaining it will be helpful.


Have you tried with `JSON.parse`?

fernandomg (commented):
couple of comments

src/apps/TxBuilder/components/Dashboard.tsx
nicosampler: yes, I did and it was not working in some cases. I didn't find another alternative than using eval .

package.json
```diff
@@ -15,13 +15,14 @@
     "big.js": "^5.2.2",
     "date-fns": "^2.16.1",
     "graphql": "^14.6.0",
+    "http-proxy-middleware": "^1.0.5",
```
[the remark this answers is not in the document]
nicosampler: I used it to do some tests with this: https://create-react-app.dev/docs/proxying-api-requests-in-development/. Then I forget to remove it. 

I decided to go with react-app-rewired, in order to allow CORS in dev mode.

fernandomg (approved):
it's working great

Something that confused me a bit was the example, I tested with a `tuple` type, not a `tuple[]`

For the following example
```solidity
contract UserDirectory {
    struct Contact {
        string email;
        string phone;
    }
    struct User {
        string name;
        address addr;
        Contact contact;
    }
    struct Admin {
        string name;
        address addr;
    };
    function addUser(User user) public {...}
    function addAdmin(Admin admin) public {...}
}
```

for `addUser` input must be: `['name', 'address',['email','phone']]`

for `addAdmin` input must be: `['email','phone']`

dasanra (changes requested):

src/apps/TxBuilder/components/Dashboard.tsx
```diff
@@ -226,155 +232,219 @@ const Dashboard = () => {
...
-
-          {/* Input ETH value */}
-          {isValueInputVisible() && (
-            <>
-              <TextField
-                style={{ marginTop: 10, marginBottom: 10 }}
-                value={value}
-                label="ETH"
-                onChange={(e) => setValue(e.target.value)}
+    <>
```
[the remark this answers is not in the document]
nicosampler: yes, because after TextField we are adding a `<br />`

src/apps/TxBuilder/components/Dashboard.tsx
mikheevm: what happens if you put ['alert('Gotcha')]

nicosampler: It will prompt the alert, of course, it's evaluating JS code.
What's **REALLY** important: It's not recognized as a valid parameter and the TX is not added to the list. ;)

nicosampler:
@fernandomg is it better now? 
Note that I also changed the input helper from `Tuple[uint256, string]` to `Tuple(uint256, string)` also added the distinction with it's an array `Tuple(uint256, string)[]`
image

src/apps/TxBuilder/components/Dashboard.tsx
rmeissner: why is it not possible to expect the array input as json syntax, so that you can do `JSON.parse(cleanValue)`?

nicosampler: This is not a valid JSON `json.parse(['1', '2'])`. 
`Uncaught ReferenceError: json is not defined`

mikheevm: @nicosampler 
it has to be JSON instead of json
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse

nicosampler: you're right, but the result it's the same.

image

nicosampler: I found a way to make it work without `eval` by escaping the `"` character: `JSON.parse(cleanValue.replace(/"/g, '"'))`

mikheevm: @nicosampler this snippet of code replaces " with ", doesn't it?

mikheevm: Screenshot 2020-09-21 at 17 04 58

nicosampler: yes! The input value provided by the UI does not scape the `"`. So we have two alternatives, request the user to scape the value like so: \\" or let the user include the raw `"` and do the work ourselves.

src/apps/TxBuilder/components/Dashboard.tsx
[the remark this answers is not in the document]
nicosampler: fixed!

dasanra (approved):

PR #33, opened 2020-09-11
Add Safe-Connect app

nicosampler:
- Refactor safe-connect app based on Tobi's repository.
- Add CSS to be aligned with Sasha design


lukasschor:
What would be the best way for me to test this? :)

nicosampler:
I'll create a temporary ipfs deploy on Monday

src/apps/SafeConnect/utils.ts
```diff
@@ -0,0 +1,21 @@
+export const blobToImageData = async (blob: string) => {
+  // let blobUrl = URL.createObjectURL(blob);
+
+  return new Promise((resolve, reject) => {
+    let img = new Image();
+    img.onload = () => resolve(img);
+    img.onerror = (err) => reject(err);
+    img.src = blob;
+  }).then((img: any) => {
```
mikheevm, line 9: IMAGE 2020-09-13 20:42:04

src/apps/Compound/config.ts
```diff
@@ -22,7 +22,12 @@ export type TokenItem = {
 export const web3Provider = process.env.REACT_APP_WEB3_PROVIDER_URL || "";
 
 export const getTokenList = (network: Networks): Array<TokenItem> => {
+  if (network !== "rinkeby" && network !== "mainnet") {
```
rmeissner, line 25: To be honest I would prefer if we clearly separate changes ... this PR should be for the Safe Connect app and here we make changes to the compound app. In general we should handle them more separately like a mono repo setup. We do something similar in the safe-modules repo where each subfolder represents a module and is independent from the others. Else we will run into situation where you update a lib and this always requires to update all Safe apps creating more changes then necessary.

src/apps/SafeConnect/App.tsx
```diff
@@ -0,0 +1,211 @@
...
+        const reader = new FileReader();
+        reader.onload = async (event: any) => {
+          const imageData = await blobToImageData(event.target.result);
+          const code = jsQr(imageData.data, imageData.width, imageData.height);
+          if (code && code.data) {
+            setIsConnecting(true);
+            wcConnect(code.data);
+          } else {
+            setInvalidQRCode(true);
+            setWcUri(
```
rmeissner, line 84: This just looks wrong ... we set some text into a field that we call "Uri"

src/apps/SafeConnect/App.tsx
```diff
@@ -0,0 +1,211 @@
...
+      <HelpLink />
+    </>
+  );
+
+  useEffect(() => {
+    let interval: number;
+    if (isConnecting) {
+      interval = setTimeout(() => setIsConnecting(false), 2_000);
+    }
+    return () => clearInterval(interval);
```
rmeissner, line 190: `clearInterval` is used with `setInterval` shouldn't we use `clearTimeout` here?

src/apps/SafeConnect/App.tsx
```diff
@@ -0,0 +1,211 @@
...
+      </StyledText>
+
+      <HelpLink />
+    </>
+  );
+
+  useEffect(() => {
+    let interval: number;
+    if (isConnecting) {
+      interval = setTimeout(() => setIsConnecting(false), 2_000);
```
rmeissner, line 188: We can add a comment here why we are doing this?

src/apps/SafeConnect/hooks/useWalletConnect.tsx
```diff
@@ -0,0 +1,86 @@
...
+    setWcClientData(null);
+  }, [connector]);
+
+  const wcConnect = useCallback(
+    async (uri: string) => {
+      const connector = new WalletConnect({
+        uri,
+        clientMeta: {
+          description: "Gnosis Safe",
+          url: "https://gnosis-safe.io/app",
```
rmeissner, line 26: should this depend on the network?

src/apps/SafeConnect/hooks/useWalletConnect.tsx
```diff
@@ -0,0 +1,86 @@
...
+      });
+      setConnector(connector);
+      setWcClientData(connector.peerMeta);
+      localStorage.setItem(LOCAL_STORAGE_URI_KEY, uri);
+
+      connector.on("session_request", (error, payload) => {
+        if (error) {
+          throw error;
+        }
+        debugger;
```
rmeissner, line 39: ?

src/apps/SafeConnect/hooks/useWalletConnect.tsx
```diff
@@ -0,0 +1,86 @@
...
+        });
+
+        setWcClientData(payload.params[0].peerMeta);
+      });
+
+      connector.on("call_request", (error, payload) => {
+        if (error) {
+          throw error;
+        }
+        debugger;
```
rmeissner, line 52: ? (if this PR is still work in progress please mark it as such)

src/apps/SafeConnect/hooks/useWalletConnect.tsx
```diff
@@ -0,0 +1,86 @@
...
+            },
+          ]);
+        }
+      });
+
+      connector.on("disconnect", (error, payload) => {
+        if (error) {
+          throw error;
+        }
+        debugger;
```
rmeissner, line 69: again

src/apps/SafeConnect/hooks/useWalletConnect.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const wcDisconnect = useCallback(async () => {
+    connector?.killSession();
+    localStorage.removeItem(LOCAL_STORAGE_URI_KEY);
+    setConnector(undefined);
+    setWcClientData(null);
+  }, [connector]);
+
+  const wcConnect = useCallback(
+    async (uri: string) => {
+      const connector = new WalletConnect({
```
rmeissner, line 22: This shadows another variable

rmeissner (changes requested):

src/apps/SafeConnect/hooks/useWalletConnect.tsx
```diff
@@ -0,0 +1,86 @@
...
+      localStorage.setItem(LOCAL_STORAGE_URI_KEY, uri);
+
+      connector.on("session_request", (error, payload) => {
+        if (error) {
+          throw error;
+        }
+        debugger;
+        connector.approveSession({
+          accounts: [safe.getSafeInfo().safeAddress],
+          chainId: safe.getSafeInfo().network === "rinkeby" ? 4 : 1,
```
rmeissner, line 42: This logic is something we need to overthing everywhere as we want to start support for "xdai" ... also in this app we do not limit it to `rinkeby` and `mainnet` (as it was adjusted for compound for exmaple)

src/apps/Compound/config.ts
nicosampler: Agreed about handle this repo as a mono repo. What do you suggest to do right now, should I extract this change in a separated PR?

src/apps/SafeConnect/App.tsx
nicosampler: sure!

src/apps/SafeConnect/hooks/useWalletConnect.tsx
nicosampler: yep

src/apps/SafeConnect/App.tsx
```diff
@@ -0,0 +1,219 @@
...
+      </StyledText>
+
+      <StyledText size="lg">
+        3) The transaction has to be confirmed be owners and executed just like
+        any other Safe transaction.
+      </StyledText>
+
+      <HelpLink />
+    </>
+  );
```
mikheevm, line 188: We've talked about defining components in another component's body many times..

mikheevm: https://twitter.com/0xca0a/status/1269562072725098501

src/apps/SafeConnect/components/WCClientInfo.tsx
```diff
@@ -0,0 +1,57 @@
...
+const FlexAlignedCenter = styled.div`
+  display: flex;
+  margin-bottom: 10px;
+
+  > :first-of-type {
+    margin-right: 5px;
+  }
+`;
+
+const Image = styled.div<{ src: string }>`
```
mikheevm, line 14: the name is misleading, this is not an image

any reason why it couldn't be a regular image?

src/apps/SafeConnect/components/WCClientInfo.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { Text, IconText, Icon } from "@gnosis.pm/safe-react-components";
+
+const FlexAlignedCenter = styled.div`
+  display: flex;
+  margin-bottom: 10px;
+
+  > :first-of-type {
+    margin-right: 5px;
+  }
+`;
```
mikheevm, line 12: there are no alignment styles but the name says so

src/apps/SafeConnect/App.tsx
```diff
@@ -0,0 +1,219 @@
...
+      for (const index in items) {
+        const item = items[index];
+
+        if (item.kind !== "file") {
+          continue;
+        }
+
+        const blob = item.getAsFile();
+        const reader = new FileReader();
+        reader.onload = async (event: any) => {
```
mikheevm, line 81: any

src/apps/SafeConnect/App.tsx
```diff
@@ -0,0 +1,219 @@
...
+);
+
+const App = () => {
+  const { wcClientData, wcConnect, wcDisconnect } = useWalletConnect();
+  const [inputValue, setInputValue] = useState("");
+  const [invalidQRCode, setInvalidQRCode] = useState(false);
+  const [isConnecting, setIsConnecting] = useState(false);
+
+  const onPaste = React.useCallback(
+    (event: any) => {
```
mikheevm, line 63: any

mikheevm (changes requested):

src/apps/SafeConnect/App.tsx
nicosampler: `getConnectedInstructions` and `getDisconnectedInstructions`, should not have problems https://twitter.com/dan_abramov/status/1269978690508787715.

`getDisconnectedContent` and `getConnectedContent` will do, what do you suggest, using React.memo or moving the component outside and pass component data through props?

src/apps/SafeConnect/components/WCClientInfo.tsx
nicosampler: 1. It's a div simulating to be an image, so the name it's fine. 
2. https://stackoverflow.com/a/1469139/2524608

src/apps/SafeConnect/components/WCClientInfo.tsx
nicosampler: yes, I removed the alignment and I forget to rename the component.

src/apps/SafeConnect/App.tsx
mikheevm: The tweet you sent says the opposite, doesn't it? 
> Blowing away the state of all components below on every render, including text inputs, *scroll position, focus,* etc.

src/apps/SafeConnect/components/WCClientInfo.tsx
mikheevm: Well that's debatable, IMO our usage falls into this category
```
Pragmatic uses of IMG
Use IMG plus alt attribute if the image is part of the content such as a logo or diagram or person (real person, not stock photo people). —sanchothefat
```

But since we still display the name so in our case using background image is fine

nicosampler: the main problem is that we don't know the image size, so using `background` with `cover`, we can ensure the image will look properly.

src/apps/SafeConnect/App.tsx
nicosampler: > The tweet you sent says the opposite, doesn't it?
As far as I understand it does not, it's only text elements, should have these problems, but I moved outside anyways.

src/apps/Compound/config.ts
rmeissner: hmmm yes I would suggest to have these changes in a different PR and then as a follow up issue (that we should create) we should convert this to mono repo. (Unless setting up the mono repo is fast then I would do that before)

package.json
```diff
@@ -4,14 +4,18 @@
   "private": true,
   "dependencies": {
     "@apollo/react-hooks": "^3.1.3",
-    "@gnosis.pm/safe-apps-sdk": "^0.1.1",
-    "@gnosis.pm/safe-react-components": "^0.1.3",
+    "@gnosis.pm/safe-apps-sdk": "https://github.com/gnosis/safe-apps-sdk.git#25a1afe",
```
rmeissner, line 7: We should if possible use a stable (on npm released) version

package.json
```diff
@@ -4,14 +4,18 @@
...
+    "@gnosis.pm/safe-react-components": "https://github.com/gnosis/safe-react-components.git#be51ca5",
```
rmeissner, line 8: same here ;)

src/apps/SafeConnect/hooks/useWalletConnect.tsx
nicosampler: Ok, I'll split it up. 
Moving to a Monorepo will take some time so it should be handled in a different issue.

package.json
nicosampler: here we can, because Mikhail release it today.

package.json
nicosampler: We could, but we are not using components in that way, it's pretty unstable so we decided to use commits references instead, to avoid releasing all the time. However, we don't release for a while, so perhaps it's a good opportunity to do it.

src/apps/SafeConnect/App.tsx
```diff
@@ -0,0 +1,221 @@
...
+`;
+
+const WCContent = styled.div`
+  min-width: 400px;
+  margin-right: 20px;
+`;
+
+const HelpLink = () => (
+  <StyledHelpLink>
+    <a
```
mikheevm, line 43: you can just put the styles from `StyledHelpLink` to `a`, can't you?

mikheevm (approved):
Thanks 👍 just one small thingy

src/apps/SafeConnect/App.tsx
nicosampler: I think I can't. display: flex should be applied to the container.

src/apps/Compound/config.ts
nicosampler: Ok, I'll split it up.
Moving to a Monorepo will take some time so it should be handled in a different issue.

PR #40, opened 2020-09-22
Update travis build

dasanra:
yarn install is executing twice. Not critical

We can update VM linux version so we start using a recent version of some programs (including python for pip and aws)


nicosampler (approved):

.travis.yml
```diff
@@ -1,9 +1,9 @@
 
 if: (branch = master) OR (branch = development) OR (tag IS present)
+dist: bionic
 language: node_js
 node_js: '12'
 before_install:
-  - yarn install --network-concurrency 1
```
nicosampler, line 6: Not sure about removing this line. If the package.json has 2 dependencies from Github (instead of npm), the installation will fail.

mikheevm: Why is it not a problem in safe-react then?

mikheevm: I don't think the concurrency is somehow related to where the dependencies are coming from

dasanra:
We are changing the repo structure. This configuration changes are not necessary any more

PR #43, opened 2020-09-28
Fix: Compound app not loading

nicosampler:
- Dependencies updates
- Set `ThemeProvider` in a separated file (index.tsx) as it was not being applied to `<Loader />` component in the previous implementation. 
- Move `initSdk()` outside the react component `CompoundWidget`


fernandomg (approved):
LGTM

PR #54, opened 2020-10-05
Add support for gs_multi_send rpc

rmeissner:
The CPK uses the `gs_multi_send` method when connected to a Safe via WalletConnect. This is currently not handled by the Safe app.


src/apps/WalletConnect/hooks/useWalletConnect.tsx
```diff
@@ -53,22 +53,35 @@ const useWalletConnect = () => {
...
+                value: txInfo.value || "0x0",
+                data: txInfo.data || "0x",
+              },
+            ]);
+            break;
+          }
+          case "gs_multi_send": {
+            const txs = payload.params as any[];
+            safe.sendTransactions(
+              // TODO: should we do some parameter validation here?
```
mikheevm, line 71: The interface does its own validation and if doesn't pass, it displays an error message. I think it's ok to leave it like this

src/apps/WalletConnect/hooks/useWalletConnect.tsx
```diff
@@ -53,22 +53,35 @@ const useWalletConnect = () => {
...
+              {
+                to: txInfo.to,
+                value: txInfo.value || "0x0",
+                data: txInfo.data || "0x",
+              },
+            ]);
+            break;
+          }
+          case "gs_multi_send": {
+            const txs = payload.params as any[];
```
mikheevm, line 69: 🙈

nicosampler: We should add this type in some generic place. 
```
interface Transaction {
    to: string;
    value: string;
    data: string;
}
```

src/apps/WalletConnect/hooks/useWalletConnect.tsx
nicosampler: does it make sense to validate that `data !== '0x' or value !== '0x0'` ?

src/apps/WalletConnect/hooks/useWalletConnect.tsx
mikheevm: It can be imported from the sdk

src/apps/WalletConnect/hooks/useWalletConnect.tsx
mikheevm: the question is if it makes sense to return there and not show the invalid tx modal

src/apps/WalletConnect/hooks/useWalletConnect.tsx
rmeissner: hmm the definition of a JsonRPC is that it is `any[]`, that's why I chose it. But we could use the Transaction type here as we expect the 3 values to be present.

rmeissner:
It seems that wallet connect takes the `clientMeta` from the index. See https://github.com/WalletConnect/walletconnect-monorepo/blob/next/packages/clients/core/src/index.ts#L116 and https://github.com/pedrouid/window-metadata/blob/master/src/index.ts therefore this doesn'T work right now ....

src/apps/WalletConnect/hooks/useWalletConnect.tsx
rmeissner: That is also why I added the todo, since we do blind casting from `any[]` to `Transaction[]` ... not a big fan of that

mikheevm: I understand your concern now and why we may want to perform validation there.
Could it be that `unknown` is a better option? We can also make a type-guard to check if it has a structure similar to the transaction object: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

rmeissner:
@nicosampler @mikheevm for the comment above: how to I update the meta tags for this react app. As we use a single application for all the apps I cannot just add the index.html (which would be normally super easy). In general it seems nicer to do this via js, since we have all our info there.

rmeissner:
I will wait for https://github.com/gnosis/safe-react-apps/issues/55, should make it easier to fix this meta info issue.

src/apps/WalletConnect/hooks/useWalletConnect.tsx
rmeissner: As this looks good I will check this out on friday :)

src/apps/WalletConnect/hooks/useWalletConnect.tsx
rmeissner: See comment below

mikheevm:
@rmeissner meissner I’m sorry this PR got overlooked, is it ready to be reviewed/merged or it needs some additional work?

mikheevm:
@rmeissner About the metadata issue: to me, it seems there's a flaw in walletconnect logic of assigning metadata because first, it tries to get it from  the DOM, and only if it fails, it uses the passed clientMeta option

The only point of failure it has is when the `window` or `window.document` property is inaccessible which isn't likely to happen in 99% and it results in:

Screenshot 2021-01-18 at 14 26 35

I think we should ask the WC team if it's done on purpose and possibly contribute to their repo.

rmeissner:
Yes, I would love if the walletconnect team could fix this, feel free to contact them :) 

The switch to separate repos for each app should be an intermediate fix, right?

Note: I cannot approve this PR as I created it ... but looks good now :)

mikheevm:
> Yes, I would love if the walletconnect team could fix this, feel free to contact them :)
> 
> The switch to separate repos for each app should be an intermediate fix, right?
> 
> Note: I cannot approve this PR as I created it ... but looks good now :)

Got a response that it was done on purpose for usage with node.js/react-native

mikheevm:
@rmeissner could you please check it once again? tweaked things a little

mikheevm (approved):

PR #58, opened 2020-10-08
Feature: TX-Builder - Make it work on xDai

### LINKED ISSUE #53: [xDai] - Transaction Builder ABI lookup
Parametrize the address to lookup for the ABI based on the connected network.

Currently, etherscan is hardcoded.

nicosampler:
Closes #53.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@apollo/react-hooks": "^3.1.3",
-    "@gnosis.pm/safe-apps-sdk": "0.4.1",
+    "@gnosis.pm/safe-apps-sdk": "https://github.com/gnosis/safe-apps-sdk.git#75232",
     "@gnosis.pm/safe-react-components": "^0.3.0",
     "@material-ui/core": "^4.9.0",
     "@rmeissner/safe-apps-react-sdk": "^0.3.4",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@apollo/react-hooks": "^3.1.3",
-    "@gnosis.pm/safe-apps-sdk": "https://github.com/gnosis/safe-apps-sdk.git#ea8dcab",
+    "@gnosis.pm/safe-apps-sdk": "0.4.2",
     "@gnosis.pm/safe-react-components": "^0.3.0",
     "@material-ui/core": "^4.9.0",
     "@rmeissner/safe-apps-react-sdk": "^0.3.4",
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@apollo/react-hooks": "^3.1.3",
-    "@gnosis.pm/safe-apps-sdk": "https://github.com/gnosis/safe-apps-sdk.git#75232",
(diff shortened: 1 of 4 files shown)


nicosampler:
@dasanra should we request a new environment? 

* https://safe-apps.dev.gnosisdev.com
* https://safe-apps.staging.gnosisdev.com
* https://apps.gnosis-safe.io

None of these environments allow us to test the TX-builder on xDai. Another alternative is to remove the env var and handle it in the code as we are doing in safe-react (src/config), the drawback of it is that we need to hardcode the URLs (infura and https://rpc.xdaichain.com).

src/apps/TxBuilder/hooks/useServices/interfaceRepository.ts
```diff
@@ -15,6 +16,24 @@ export interface ContractInterface {
 
 const getContract = memoize(async (apiUrl: string) => axios.get(apiUrl));
 
+const getAbiUrlByNetwork: {
```
fernandomg, line 19: the functionality is there, I'm not so sure about the `get` part of the name for this map-like object.

nicosampler: that's right, I thought it as a function then moved it into an object. any suggestion? 
removing the `get` perhaps?

fernandomg: Yes.

In fact, what this returns is not an `abiUrlByNetwork` but a function that returns the `abiUrl`.

So, thinking in that direction I thought something around:
```typescript
const getAbiUrl = abiUrlGetterByNetwork[networkName]

if (getAbiUrl === null) {
  throw Error ('...')
}

const abi = getAbi(getAbiUrl(contractAddress))
```

would be most suitable.

But I'm not the master-on-naming-things :mage_man:, so, your call.

nicosampler: done! thx for the detailed feedback :)

fernandomg (approved):

package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@apollo/react-hooks": "^3.1.3",
-    "@gnosis.pm/safe-apps-sdk": "0.4.1",
+    "@gnosis.pm/safe-apps-sdk": "https://github.com/gnosis/safe-apps-sdk.git#ea8dcab",
```
dasanra, line 7: v0.4.2 is being released right now with the necessary changes. I would use that version.

src/apps/TxBuilder/hooks/useServices/interfaceRepository.ts
```diff
@@ -25,19 +44,17 @@ class InterfaceRepository {
   }
 
   private async _loadAbiFromEtherscan(address: string): Promise<string> {
```
dasanra, line 46: `_loadAbiFromBlockExplorer`?

package.json
nicosampler: it's not yet on npm

src/apps/TxBuilder/hooks/useServices/interfaceRepository.ts
nicosampler: done!

package.json
nicosampler: @dasanra done!

dasanra (approved):

PR #59, opened 2020-10-16
Add travis deployment for xDai

dasanra:
Add instruction for apps deployment on xDai network

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 
-if: (branch = master) OR (branch = development) OR (tag IS present) OR (branch = xdai-deployment)
+if: (branch = master) OR (branch = development) OR (tag IS present)
 language: node_js
 node_js: '12'
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -45,7 +50,9 @@ deploy:
     local_dir: build
     region: $AWS_REGION
     on:
-      branch: master
+      # branch: master
+      all_branches: true
+      condition: $TRAVIS_BRANCH =~ ^(master|feature/xdai-deployment)$
```
rmeissner, line 55: here `master` is on the same level as `feature/xdai-deployment` ... in this case I would recommend not to call it `feature/...` (maybe just `xdai-deployment` or `release/...` not sure). Feature branches are for me used when developing a feature.

nicosampler: I think it's just temporal, then all the environments will be deployed together when we merge to master.

nicosampler: @rmeissner updated. I think `xdai-deployment` it's self explanatory

dasanra: Yes, this was a special case to make it deploy to staging without merging to master. I think that this can be removed now

nicosampler (approved):

Issue #56, opened 2020-10-05
[WalletConnect] Wrong client info used


rmeissner:
Currently the WalletConnect safe app uses the meta information from the root [index.html](https://github.com/gnosis/safe-react-apps/blob/development/public/index.html). This is because the wallet connect sdk tries to parse them. 

See https://github.com/gnosis/safe-react-apps/pull/54#issuecomment-703681329

# Expected

Name: Gnosis Safe Multisig
Logo: 
![image](https://user-images.githubusercontent.com/1380529/99513767-42a16b00-298b-11eb-9e48-dfa0e09ddb4f.png) (That's just a screenshot, I think the logo should be in the project already. If not, please let me know.)

# Updates

- Nov 18, Tobi: Added expected behavior

anxolin:
Yep, interested in this one since we use the returned image is broken:

![image](https://user-images.githubusercontent.com/2352112/95455269-e3a10d00-096d-11eb-8211-0d7f2495d4ae.png)

This is the information that's being sent:
![image](https://user-images.githubusercontent.com/2352112/95455326-fadffa80-096d-11eb-942d-437b529f9be4.png)

tschubotz:
Requires to refactor the way we do Safe apps. (gnosis/safe-react-apps#55)

tschubotz:
This will most likely be resolved once the gnosis/safe-react-apps#55 issue is released -> to check

nicosampler:
this should be fixed @tschubotz @rmeissner could you re-check? And close if everything is ok.

tschubotz:
@nicosampler this is how it looks now: 
![image](https://user-images.githubusercontent.com/1380529/99513118-9495c100-298a-11eb-9eea-746d063c9445.png)

Seems to work, except we display the WalletConnect logo. I think we should return the Gnosis Safe logo though.
Also as name I would use just "Gnosis Safe Multisig" instead of "Gnosis Safe Multisig WalletConect".

We should have made this more clear in the issue description what the expected behavior is, sorry.
I'll add this now.

dasanra:
@tschubotz what do you think about this?

![Captura de pantalla de 2020-11-19 09-46-51.png](https://images.zenhubusercontent.com/5f030e579c03aa152acf03e2/a5fa55e2-d075-4af4-a16e-80b1a5067141)

tschubotz:
> @tschubotz what do you think about this?

Looks perfect to me!

dasanra:
[...]

PR #69, opened 2020-11-05
Separate apps

### LINKED ISSUE #55: Separate Safe apps
Currently all Safe apps are 1 React app. This cause issues when updating dependencies and also with meta information files. Therefore we should properly separate them so that each safe app is their own react app (they can still all exist in the same repo).

nicosampler:
closes #55 #56.  
Now, each safe-app is an independent app.

- I have used `yarn workspaces` to unify dependencies.
- In the main `package.json` there are scripts to start and build each app independently.
- I have used `dotenv-cli` to only have one .env shared for all the apps. 
- I changed the logic to form the URL provider, for xdai, volta, and EWC, the URL is harcoded, for rinkeby and mainnet it uses the env var REACT_APP_RPC_TOKEN to form the URL. This will be handy to make work the same build in all networks.


rmeissner (approved):
As far as I can tell it looks good ... if I have a chance I will try all the projects locally, but won't be able to this week :/

mikheevm:
wdyt about including apps in a single folder?
```
packages/apps
  - compound
  - walletconnect
  - ...
```

this way we won't have to add a new entry to `package.json` when adding a new app

also, I'm not sure about scripts to run the apps, usually with workspaces it looks like that `yarn workspace compound run start`

mikheevm:
Also please update readme :)

nicosampler:
> wdyt about including apps in a single folder?
> 
> ```
> packages (or apps)
>   - compound
>   - walletconnect
>   - ...
> ```
> 
> this way we won't have to add a new entry to `package.json` when adding a new app

As far I understand it will not work with yarn workspaces.

> 
> also, I'm not sure about scripts to run the apps, usually with workspaces it looks like that `yarn workspace compound run start`

It's great, didn't know about it.

mikheevm:
> As far I understand it will not work with yarn workspaces.

It will.

```
  "workspaces": {
    "packages": [
      "packages/*"
    ]
  },
```

mikheevm:
We also can remove tsconfig.json from each app and use a single file, an example of such tsconfig:
```
  
{
  "compilerOptions": {
    "target": "es5",
    "lib": ["dom", "dom.iterable", "esnext"],
    "allowJs": true,
    "skipLibCheck": true,
    "esModuleInterop": true,
    "allowSyntheticDefaultImports": true,
    "strict": true,
    "strictNullChecks": true,
    "forceConsistentCasingInFileNames": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "noEmit": false,
    "jsx": "react"
  },
  "include": ["./apps/*/**/src"],
  "exclude": ["dist", "node_modules"]
}
```

nicosampler:
@mikheevm I tested the single tsconfig.json for all the apps, but it's not working, typescript automatically creates a `tsconfig` on each app when you start the app, and the first time you run it without the tsconfig it fails, if you re-run the app, now with the tsconfig created automatically, it works fine. 

I have been looking at other projects and all of them have a file for each workspace.

dasanra (commented):
As we talked. It would be nice to remove Storybook dependencies as they aren't in use

nicosampler:
@dasanra I just pushed a commit that removed storybook.

dasanra (approved):

dasanra:
@nicosampler I scheduled a meeting with devops to comment on how to handle automatic build and deployment process. Until then we can handle them as external apps

dasanra:
Added a comment in the issue on why we are marking this as resolved. Only CI is pending

https://github.com/gnosis/safe-react-apps/issues/55#issuecomment-726112149

PR #70, opened 2020-11-09
wallet-connect: accept uri as pasted value.

### LINKED ISSUE #57: [WalletConnect] Cannot submit pasted url
# Context

If you paste the text url (e.g. from the console on https://example.walletconnect.org/) you cannot continue.

# Task

- Allow pasting a WalletConnect URI
- Let's please keep the image paste as well.

nicosampler:
closes #57.


nicosampler:
this can be tested here connection to this site https://example.walletconnect.org/. Just copy and paste the value below the QR code.

dasanra (commented):
I have some doubts regarding this function.

I see that if the item is type string  and it's a wallet connect valid string it will connect but will continue the for loops checking more clipboard items?

If I copy several wallet connect links then will it connect to the first one I copied or to the last one?

Also, I'm not really into the use of loop break statements unless strictly necessary. Couldn't we here just filter inside an IF the functionality checking that `item.kind` is a string or either a file first?
And then if it is `text/plain` we would use string connect block or we fallback to file connect block

dasanra (approved):
Remember to change tests adding that now it's also able to use wallet connect link apart from QR code

dasanra:
A version of this WalletConnect app is deployed here:
https://ipfs.io/ipfs/QmWjqH6nYpRVFyqQowafsjcbayz8PZ6j1uQcLYGz5ddhzB/

dasanra:
This closes #72 too as we were adding text changes here

francovenica:
Used the "Copy to Clipboard" option that WC example gives you, and pasted it in the new WC app and it connected just fine
I tried it with the older version of the WC and it won't work there (just to be sure that this wasn't working withouth this upgrade)

Looks good to me

PR #77, opened 2020-11-18
Add GlobalStyles in Tx-builder and WC

### LINKED ISSUE #76: Header in WalletConnect Safe app has wrong font
Since the last WalletConnect Safe app update, the header font seems to be wrong.

https://user-images.githubusercontent.com/1380529/99513572-14bc2680-298b-11eb-8a98-b34570b828ab.png

I'm assigning "medium" just because we recently switched to the ipfs deployment for Safe apps and so we can check out if this is a bigger issue or not. I assume the same will happen once we move the other Safe apps to ipfs as well.

nicosampler:
Closes #76.

When safe-apps were separated on independent projects, we forgot to add GlobalStyles in tx-builder and wallet-connect. That's why wallet-connect is showing the wrong font.

### DIFF
--- a/apps/tx-builder/src/global.ts
+++ b/apps/tx-builder/src/global.ts
@@ -1,0 +1,28 @@
+import { createGlobalStyle } from "styled-components";
+import avertaFont from "@gnosis.pm/safe-react-components/dist/fonts/averta-normal.woff2";
+import avertaBoldFont from "@gnosis.pm/safe-react-components/dist/fonts/averta-bold.woff2";
+
+const GlobalStyle = createGlobalStyle`
+    html {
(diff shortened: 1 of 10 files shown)


dasanra (approved):

dasanra:
App can be tested here:

https://ipfs.io/ipfs/QmayQ1dieH4mXspYKd7V53HQBsxAZxHtkSLU96Wopi78fx

francovenica:
Didn't noticed the title with "Times new roman" font.
I added the new app given by Dani and it looks good now

Tested here: https://rinkeby.gnosis-safe.io/app/#/
Before: 
image

After: 
image

Issue #74, opened 2020-11-16
Transaction builder: Don't display events


liliya-soroka:
### Title/Description
The events are displayed together with the functions

### Environment
 - Browser: Chrome
 - Safe: 0x1230B3d59858296A31053C1b8562Ecf89A2f888b ( any)
 - Wallet: MM
 - App Version: 2.14.1
 - Environment:
   - staging (rinkeby)

### Steps to reproduce
1. Go to Transaction builder
2. Add contract address `0xc778417e063141139fce010982780140aa0cd5ab`
3. Check functions list
Current result: There are both the functions and the events in the list
![image](https://user-images.githubusercontent.com/338622/99232115-903e9c00-2802-11eb-9481-3df612455a95.png)

### Expected result
Only functions should be displayed

tschubotz:
@liliya-soroka thanks for reporting. I just transferred the issue to the safe-react-apps repo. All bugs/issues inside safe apps should go there :)

fernandomg:
closed in #79

PR #79, opened 2020-11-20
(Fix) - No eth field for payable functions on tx builder

### LINKED ISSUE #75: Transaction builder: the app is crashed on the "empty" line in the functions list
### Title/Description
The Transaction builder app is crashed with the error "Cannot read property 'map' of undefined" in the console when "empty" ( assuming Fallback function) is selected for the contract `0xc778417e063141139fce010982780140aa0cd5ab`

### Environment
 - Browser: Chrome
 - Safe: any
 - Wallet: MM
 - App Version: 2.14.1
 - Environment:
   - staging (rinkeby)

### Steps to reproduce
1. Go to the Transaction builder
2. Enter contract address `0xc778417e063141139fce010982780140aa0cd5ab`
3. Select "empty" function under deposit
Current result: The app is crashed

### Expected result
The app should not be crashed.

### Screenshots
image

Agupane:
Closes #75 and #74 by:

- Excluding from the abi the methods that aren't of type `function`

Also fixes [#1605](https://github.com/gnosis/safe-react/issues/1607) by:

- Checking that the method is `payable` or not in order to enable the `ETH` input


apps/tx-builder/src/hooks/useServices/interfaceRepository.ts
```diff
@@ -72,6 +72,10 @@ class InterfaceRepository {
           return false;
         }
 
+        if(e?.type.toLowerCase() !== 'function') {
```
rmeissner, line 75: We should also support `fallback` and `receive` functions (see https://docs.soliditylang.org/en/v0.7.5/abi-spec.html#json)

rmeissner: This should also impact the eth field

Agupane: Thanks for the feedback! I improved it now

apps/tx-builder/src/hooks/useServices/interfaceRepository.ts
```diff
@@ -72,9 +72,16 @@ class InterfaceRepository {
...
 
+        if(e?.type.toLowerCase() === 'event') {
+          return false
+        }
+
         return !e.constant;
       })
       .map((m: any) => {
+        if(m.type === 'fallback' || m.type === 'receive') {
+          return { inputs: [], name: m.type === 'fallback'  ? 'fallback' : 'receive', payable: m.payable || false };
```
rmeissner, line 83: Could we simplify `m.type === 'fallback'  ? 'fallback' : 'receive'` to `m.type`, as we check before that it is either one of those two strings?

rmeissner (approved):

apps/tx-builder/src/hooks/useServices/interfaceRepository.ts
Agupane: I improved it, please check it

rmeissner: looks good

dasanra (approved):

nicosampler (approved):

PR #85, opened 2021-01-11
Feature: Update apps to V1 of the SDK

mikheevm:
This PR:
- Updates SDK version used in apps to v1
- Removes duplicated dependencies across apps
- Some linting/code style fixes

## Misc
Changes in tsconfig.json are generated by react-script-rewired (it generates/updates tsconfig.json every launch/build)


apps/compound/src/config.ts
```diff
@@ -1,14 +1,14 @@
-import { Networks } from "@gnosis.pm/safe-apps-sdk";
+import { Networks } from '@gnosis.pm/safe-apps-sdk';
```
mikheevm, line 1: This file contains only linting changes.

apps/compound/src/abis/CWEth.ts
```diff
@@ -1,596 +1,598 @@
-export default [
+const CWethABI = [
```
mikheevm, line 1: This file contains only linting changes.

apps/compound/src/abis/CToken.ts
```diff
@@ -1,1124 +1,1126 @@
-export default [
+const CTokenABI = [
```
mikheevm, line 1: This file contains only linting changes.

apps/compound/src/abis/CErc20.ts
```diff
@@ -1,617 +1,619 @@
-export default [
+const CErc20ABI = [
```
mikheevm, line 1: This file contains only linting changes.

apps/compound/src/tokens.ts
```diff
@@ -1,55 +1,57 @@
 // reference
```
mikheevm, line 1: This file contains only linting changes.

apps/compound/src/App.tsx
```diff
@@ -1,84 +1,74 @@
-import React, { useEffect, useState } from "react";
```
dasanra, line 1: This file is having semi-colons and single quote

apps/compound/src/abis/CErc20.ts
```diff
@@ -1,617 +1,619 @@
-export default [
```
dasanra, line 1: This file is having single quotes and semi-colons

apps/compound/src/index.tsx
```diff
@@ -1,25 +1,28 @@
-import React from "react";
```
dasanra, line 1: This file is having double quote but no semi-colonos (it's a .tsx)

apps/compound/src/abis/CToken.ts
dasanra, line 1: This file is having single quote and semi-colons (it's a .ts)

apps/compound/src/abis/CWEth.ts
dasanra, line 1: This file is having single quotes and semi-colons

apps/compound/src/config.ts
```diff
@@ -1,14 +1,14 @@
-import { Networks } from "@gnosis.pm/safe-apps-sdk";
```
dasanra, line 1: This file is having single quotes and semi-colons

apps/tx-builder/src/components/Dashboard.tsx
```diff
@@ -7,17 +7,18 @@ import {
   Select,
```
dasanra, line 7: I think there should be some wrong configuration of the linter as .ts and .tsx files use different linting

package.json
```diff
@@ -10,10 +9,41 @@
     "build-tx-builder": "dotenv -e .env -- yarn workspace tx-builder build",
     "build-wallet-connect": "dotenv -e .env -- yarn workspace wallet-connect build"
   },
+  "version": "2.0.0",
   "workspaces": [
     "apps/*"
   ],
+  "dependencies": {
```
dasanra, line 16: I don't have an strong opinion on devDependencies, as they aren't packed when compiling the app.

I see that this dependencies are used in the three apps, but if we do this we should be careful when adding new apps. If an app is not using react, react will be packed anyway.

This also can add difficulties when upgrading dependencies and breaking changes. If a dependency upgrade is critical for one app but we have breaking changes with other app we should shift again to have dependencies at the app level if we want to deliver a hotfix without upgrading both apps.

Just wanted to comment this because that were the reasons of leaving it like that before.

dasanra (changes requested):
Just making sure that the linting lines follow the same guidelines

package.json
mikheevm: >  If an app is not using react, react will be packed anyway.

I don't think so, the bundler will only include what's imported in the app.

> This also can add difficulties when upgrading dependencies and breaking changes. If a dependency upgrade is critical for one app but we have breaking changes with other app we should shift again to have dependencies at the app level if we want to deliver a hotfix without upgrading both apps.

True. But apps are pretty small, and, for example, updating the SDK for every app was pretty much changing only a few lines of code. Adding a dependency on the app level (for the case you mentioned) is not that hard (just running 1 command basically), so I think it's fine. React.js rarely introduces breaking changes. I did the change mostly to motivate us to keep the dependencies up-to-date 😄

apps/tx-builder/src/components/Dashboard.tsx
mikheevm: Seems to be the same (single quote and semi-colons), I just haven't applied it to some files

mikheevm: Previously the projects used different configuration, now it's one configuration but I haven't formatted/linted the whole project so there can be some inconsistencies

package.json
rmeissner: This was a requirement by me. Updating dependencies for all apps at the same time broke functionality before.

If I remember correctly we now publish everything via ipfs so it is less critical, but in general it should be possible to update/ add dependencies for one of the Safe apps without touching the other Safe apps.

mikheevm: It's still possible. Only sdk, react and styled components are shared. But locking a dependency for one app is as easy as running `yarn workspaces *app-name* add dep@v1`

apps/tx-builder/src/components/Dashboard.tsx
dasanra: Yes, I'm fine to start linting modified files, but I was confused to see different rules applied to some files where the linter clearly run

dasanra (approved):

mikheevm:
@dasanra do we wait for QA to review or they will review the release candidate?

dasanra:
@mikheevm I think we can go ahead. Apps will be tested for release v2.19 and we have to deploy them to IPFS in order to make Franco able to test. We can do QA in functionality PR #86

PR #89, opened 2021-01-14
Feature: eth_sign

### LINKED ISSUE #87: [WalletConnect] handle `eth_sign`
Safe apps currently don't have any way to generate pure off-chain signatures (that don't require a transaction) that can be validated via EIP-1271. BUT the safe has a method called `signMessage` that can be used to mark a message hash as "signed" so that when the EIP-1271 callback is called it is a valid signature.

Steps:
* the hash passed via `eth_sign` needs to be converted to bytes and a transaction that calls `signMessage` with it needs to be created (see https://github.com/gnosis/safe-contracts/blob/development/contracts/GnosisSafe.sol#L307) 
* A signature with 0 length needs to be returned to the dapp (see https://github.com/gnosis/safe-contracts/blob/development/contracts/GnosisSafe.sol#L330)
* The dapp needs to call the *old* EIP-712 validation method (`isValidSignature(bytes calldata _data, bytes calldata _signature)`)
  * For rinkeby we can write a new fallback handler that converts the new method to the old method

mikheevm:
This PR:
- Closes #87 
- Adds `eth_sign` handler
- Splits utils by purpose
- Small dependency bump


mikheevm:
@rmeissner could you please check if anything is missing?

apps/wallet-connect/src/hooks/useWalletConnect.tsx
```diff
@@ -48,22 +49,75 @@ const useWalletConnect = () => {
...
+                  to: safe.safeAddress,
+                  value: '0x0',
+                  data: callData,
+                },
+              ],
+            });
+
+            wcConnector.approveRequest({
+              id: payload.id,
+              result: '',
```
rmeissner, line 83: maybe we should return `0x` 🤔 on low level ethereum representation that is more correct

apps/wallet-connect/src/utils/networks.ts
```diff
@@ -0,0 +1,31 @@
...
+  UNKNOWN: 1337,
+};
+
+const gnosisUrlByNetwork: {
+  [key in Networks]: null | string;
+} = {
+  MAINNET: 'https://gnosis-safe.io/app/#',
+  MORDEN: null,
+  ROPSTEN: null,
+  RINKEBY: 'https://safe-team-rinkeby.staging.gnosisdev.com/app/#',
```
rmeissner, line 22: why do we use staging here (I know is not from this issue, but we do have a rinkeby prod app)

rmeissner (approved):

PR #92, opened 2021-01-20
Feature: Github actions (CI, test, lint)

### LINKED ISSUE #91: Update CI setup
After migration to a monorepo, CI setup got removed. This issue is to create a new setup with the following:

- Development branch is deployed to its own environment
- PRs are deployed
- Master branch is deployed to a staging environment
- Productions deployments are manual

mikheevm:
This PR:
- Closes #91 
- Adds CI for PR deployments, dev/staging envs, testing, linting, CLA check
- Removes usage of a global .env file in favor of file local per app

**Todo later in a different PR:**
- Only build/deploy changed apps


.github/workflows/deploy.pr.yml
```diff
@@ -0,0 +1,63 @@
...
+          sudo apt-get -y install python3-pip python3-dev
+          pip install awscli --upgrade --user
+          # If you have dependencies from git (wc app), yarn may randomly throw an error about invalid cache
+          # This approach is taken from https://github.com/yarnpkg/yarn/issues/7212#issuecomment-506155894 to fix the issue
+          # Another approach is to install with flag --network-concurrency 1, but this will make the installation pretty slow (default value is 8)
+          mkdir .yarncache
+          yarn install --cache-folder ./.yarncache --frozen-lockfile
+          rm -rf .yarncache
+          yarn cache clean
+      - uses: actions/cache@v2
```
dasanra, line 37: I have one doubt here. Why are you calling cache action after `yarn build` ?

mikheevm: It's yarn install, isn't it?

dasanra: Yes, I mean `yarn install`.

But I'm talking about the action used below. `actions/cache@v2`
Using it like that you are doing a full yarn install and then if there is a valid cache from previous build copying the files again

mikheevm: But it caches node_modules, not the yarn cache?

mikheevm: Oh, you mean that it should be before yarn install, don't you? I'm not sure. How will it cache files if they're not there yet? I think they are available inside the job every run by default, so you don't need to run it beforehand

dasanra: Ok, so I thought you found an special reason. ;)
That action first looks if there is something cached and try to load it. If node_modules has the expected dependencies versions (that it should because we are using yarn.lock in cache hash to check if it's compatible) `yarn install` would be way faster, so repeating builds will take less time

That action adds before Github clears everything up the "backup cache" step where it effectively stores the files to cache

mikheevm: @dasanra please check again, thanks for catching it

mikheevm: with the proper setup it shaved almost ~2m from the ci job, nice :)

dasanra (approved):

PR #115, opened 2021-04-01
Tx-builder: Refactor code and fixes #108

### LINKED ISSUE #108: [Transaction Builder] Cannot send eth to an EOA address and fallback is wrongly encoded
1. Currently it is not possible to send ETH to a non-contract (e.g. EOA) address. As the tx builder will just fail when trying to load the ABI. A solution would be to check if there is code at the address and if not allow to send eth.

2. Another issue related to sending ETH is that when interacting with the fallback method a data payload is generated.

image

It would be exected that for the `fallback` and `receive` method no payload is generated.

nicosampler:
This PR closes #108.

The code for tx-builder was a little messy, it was a huge file with all the logic, so I decided to split the code up into several files/components. 

I think the code still needs another refactor, to simplify how the txs are being validated and created but I didn't want to add more complexity to this PR. 

The fixes for #108 consisted of two parts.

1. When to make eth input visible
https://github.com/gnosis/safe-react-apps/pull/115/files#diff-16e5843f84a6e1b9db9232230df37ba17898969b8f316153ba26767187885567R191

I added a `useEffect` that determines when eth value is visible. Basically when the address is not a contract or when the contract has a payable method.

2. Avoid creating tx data when the method is receive or fallback.
https://github.com/gnosis/safe-react-apps/pull/115/files#diff-16e5843f84a6e1b9db9232230df37ba17898969b8f316153ba26767187885567R129

*. Besides splitting up the code, I changed how the `to` value is set. 
When the user enters a valid address in the main input, the same value is copied `toInput` state (the controlled value to To input). 
https://github.com/gnosis/safe-react-apps/pull/115/files#diff-16e5843f84a6e1b9db9232230df37ba17898969b8f316153ba26767187885567R185


.github/workflows/deploy.pr.yml
```diff
@@ -9,11 +9,15 @@ env:
 jobs:
   deploy:
     name: Deploy Pull Request
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
```
rmeissner, line 12: Is this required for this PR? If so can we add it to the readme? ;)

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+
+  const handleInput = useCallback(
+    async (inputIndex: number, input: string) => {
+      inputCache[inputIndex] = input;
+      setInputCache(inputCache.slice());
+    },
+    [inputCache],
+  );
+
+  const getInputHelper = (input: any) => {
```
rmeissner, line 81: A lot of the functions here don't use react hooks, should we fix this?

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+  // set when inputValue is visible
+  useEffect(() => {
+    const isVisible = async () => {
+      if (contract) {
+        const method = getContractMethod();
+        setIsValueInputVisible(method?.payable || false);
+      } else {
+        try {
+          const code = await services?.web3?.eth.getCode(toInput);
+          setIsValueInputVisible(code === '0x');
```
rmeissner, line 199: Why are we doing this? Couldn't we just allow it when we don't have an abi? We cannot really say if it is possible or not to send eth ... not sure if defaulting to "not allowing" it for contracts is the best choice.

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+        if (i > 0) {
+          description += ', ';
+        }
+        const input = method.inputs[i];
+        description += (input.name || input.type) + ': ' + cleanValue;
+      }
+      description += ')';
+
+      try {
+        if (!['receive', 'fallback'].includes(method.name)) {
```
rmeissner, line 129: Should we skip the whole description/ method parsing (so have this check at the beginning of the if in line 113) and directly go the the next part?

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+      }
+    }
+
+    try {
+      const cleanTo = web3.utils.toChecksumAddress(toInput);
+      const cleanValue = valueInput.length > 0 ? web3.utils.toWei(valueInput) : 0;
+      if (data.length === 0) {
+        data = '0x';
+      }
+      if (description.length === 0) {
```
rmeissner, line 144: Shouldn't we rather check if data is 0?

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+      transactions.push({
+        description,
+        raw: { to: cleanTo, value: cleanValue, data },
+      });
+      setInputCache([]);
+      setTransactions(transactions);
+      setSelectedMethodIndex(0);
+      setValueInput('');
+    } catch (e) {
+      console.error(e);
```
rmeissner, line 156: Shouldn't we display an error here?

rmeissner (commented):
We have some logic in the component, that could we extracted:
- Building descriptions
- Building "raw" tx for Safe Apps SDK

Different question:
Would it make sense to split this in multiple PRs?
1. Dependency updates
2. Refactoring
3. Fixing #108

.github/workflows/deploy.pr.yml
dasanra: I changed this because GithubActions wasn't working properly yesterday. Since last week ubuntu-latest uses Ubuntu 20 instead of Ubuntu 18, but I think that it was a general failure in GithubActions

apps/tx-builder/src/components/Builder.tsx
nicosampler: makes sense

apps/tx-builder/src/components/Builder.tsx
nicosampler: good catch

apps/tx-builder/src/components/Builder.tsx
nicosampler: I think it will give the same result, but checking for `data` it's more clear.

nicosampler:
> We have some logic in the component, that could we extracted:
> Building descriptions
> Building "raw" tx for Safe Apps SDK

Yes, I mentioned that in the PR description, Also I would add how we handle the errors, actually, we have one error for all the inputs, I would change that so each input has its own error, TextField input already supports that. 

> Different question:
> Would it make sense to split this in multiple PRs?

the fix for #108 it's really tiny, just a few lines, perhaps we can create a different issue/PR to continue refactoring the points mentioned above.

apps/tx-builder/src/components/Builder.tsx
nicosampler: mmm we could but I think is not really necessary, most of these handlers would be re-created anyway because some of the main inputs changed (address or abi, dropdown for methods)

apps/tx-builder/src/components/Builder.tsx
nicosampler: Added it.
I just realized we were allowing enter an Eth value lower than 0. The proper validation was added.

.github/workflows/deploy.pr.yml
dasanra: Now, github actions changes are kept to the minimum, just to fix small issues

apps/tx-builder/src/components/Builder.tsx
nicosampler: So.. if we wrap it in a `useCallback`, the dependencies array will be empty and it will behave as it is now. 
We could use `React.useMemo`, but does not make sense either. 

Are you thinking of another alternative? I would use those wrappers only if we found a performance issue

rmeissner: if the dependencies callback is empty it will only be recreated if the component is recreated ... currently it is recreated on each render.

rmeissner: useCallback is the same as useMemo. See https://reactjs.org/docs/hooks-reference.html#usecallback

nicosampler: the change you suggest is not recommended, because, in the end, there is not performance.
https://kentcdodds.com/blog/usememo-and-usecallback

But I can do it anyways if you want

rmeissner: Ok, lets take it one step further, this method doesn't need to be inside the component at all. It could be declared as a constant function outside of this component.

nicosampler: agreed on that

dasanra:
We shouldn't leave this PR behind. I was trying to validate it and I see that the changes to solve #108 are really small, but I totally get that this goes against our statement of keeping PRs simple and focused on the task they should solve. It's difficult now to trust that the most of the changes are just to keep the code as it was but splitting it into different files, because it's difficult to find the lines of the actual fix.

I was trying to validate this fix as it is. One thing I detected is that if you add one transaction and then you delete it while reviewing the list, a magic 0 appears on the interface

nicosampler:
@dasanra I can try to create a PR with only the fix, it should be pretty simple. I think it's only 2 lines of code.

But we also should also review/merge this PR.

wdyt is best?

francovenica:
The from the URL's only the Tx builder loads. the WC and Compound Apps keep loading forever

I used the dev environment to load my safe and load the apps there as "custom apps". Only the Compound app loads properly and can be used to create a transaction, the other 2 just load forever

https://safe-team.dev.gnosisdev.com/app/#/safes/0x9913B9180C20C6b0F21B6480c84422F6ebc4B808/transactions

dasanra:
@francovenica For me all three are working correctly. When you open the app links directly is normal that those 2 only show a loader, because they are waiting to be inside the Safe iframe.

I could add any of the three apps without issues in dev environment :thinking:

francovenica:
Safe and env: https://safe-team.dev.gnosisdev.com/app/#/safes/0x9913B9180C20C6b0F21B6480c84422F6ebc4B808/balances

My bad. I was copying the URL as text and the custom app seems to add a default "http://" instead of "https://" and it was causing an error (or at least is the only explanation I can find since I just copy/pasted the urls :shrugs:)
I was able to use now the WalletConnect app to create and execute a tx

I could access the Tx builder, but I could not use the tx builder:
I cannot write anything in the "Address" input. I cannot type or copypaste anything in there
I tried to use the Transfer method of the ERC20 ABI, to transfer an ERC20 from the safe to some other account
In the address field I need to enter the address of a token I have, but since I cannot enter anything I cannot use the Tx builder

ERC20 ABI: 
https://ethereumdev.io/abi-for-erc20-contract-on-ethereum/

Gif:

nicosampler:
@francovenica it should be fixed now

francovenica:
@nicosampler Ok, the issue with the Address field was solved. But I have another issue now

I complete the entire form with valid values, but I cannot add transactions. When I press the button the tx is not being added at all.
I tested this same values in the other Tx builder (The one that is there by default) and it works fine.
I don't see errors in the console or in the network tab when I click the button

To reproduce this actually there is no need to fill the form. You can compare it using both txBuilders and just hitting the Add transaction button. (you can use the safe address hash in the ABI field to show the full form)

The Tx builder by default will show you an error because the fields are empty
The Tx builder added as a custom app will not show the error
Is like if the Add Transaction button had no onClick action at all

nicosampler:
@francovenica yes, sorry 🤦‍♂️ , seems like I didn't test it yesterday.
It should be working now!

francovenica:
Is working fine now. Approved

francovenica (approved):

dasanra (approved):

PR #117, opened 2021-04-05
 Transaction Builder - Change receive method to constructor

### LINKED ISSUE #109: [Transaction Builder] Constructor is rendered as receive
When interacting with the Safe proxy (e.g. use the address of the Safe as the target in the builder) a `receive` method is shown with a parameter `_masterCopy`. 

There are 2 issues with that:
1. The real method would be the `constructor` (looking at the ABI response from etherscan), which is not interactable on deployed contracts.
2. The `receive` method cannot have any parameter and can only receive ETH

matextrem:


apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    sdk.txs.send({ txs: transactions.map((d) => d.raw) }).catch(console.error);
+  }, [sdk, transactions]);
+
+  // Set toInput when to changes
+  useEffect(() => {
+    const value = isValidAddress(to) ? (to as string) : '';
+    setToInput(value);
+  }, [to, isValidAddress]);
```
mikheevm, line 188: this is an infinite loop, isn't it?

matextrem: Why? `SetToInput` does not change any dependency in the array (to, isValidAddress)

rmeissner (changes requested):
Could we please comment what is all changed in this PR?

The issue linked just described that the `constructor` is not recognized. Looking at this PR we change the CI, the react components, add new UI components. This seems like a bigger refactoring. While this is ok, it would be nice to mention what is being done and why it is done.

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+
+  const handleInput = useCallback(
+    async (inputIndex: number, input: string) => {
+      inputCache[inputIndex] = input;
+      setInputCache(inputCache.slice());
+    },
+    [inputCache],
+  );
+
+  const getInputHelper = (input: any) => {
```
rmeissner, line 81: Why do we not use any react hooks here? (E.g. useCallback). Also why is this `any`?

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+        input.type.endsWith('[]') ? '[]' : ''
+      }`;
+    } else {
+      return input.type;
+    }
+  };
+
+  const getContractMethod = useCallback(() => contract?.methods[selectedMethodIndex], [contract, selectedMethodIndex]);
+
+  const handleSubmit = () => {
```
rmeissner, line 94: shouldn't this also be `useCallback`

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+    sendTransactions();
+    setTransactions([]);
+    setReviewing(false);
+  };
+
+  const handleDismiss = () => {
```
rmeissner, line 100: Same here

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+    setReviewing(false);
+  };
+
+  const addTransaction = async () => {
```
rmeissner, line 104: And here

rmeissner (changes requested):

apps/tx-builder/src/hooks/useServices/interfaceRepository.ts
```diff
@@ -76,7 +75,7 @@ class InterfaceRepository {
       .map((m: any) => {
         return {
           inputs: m.inputs || [],
-          name: m.name || (m.type === 'fallback' ? 'fallback' : 'receive'),
+          name: m.name || (m.type === 'fallback' ? 'fallback' : m.type === 'constructor' ? 'constructor' : 'receive'),
```
rmeissner, line 78: Maybe we should first filter by known types. Currently we filter by "unwanted" types (e.g. event). If now in a future solidity version a new type is added we will run into a similar issue. Also the contructor should never be rendered to the user.

apps/tx-builder/src/hooks/useServices/interfaceRepository.ts
```diff
@@ -76,7 +75,7 @@ class InterfaceRepository {
...
           payable: this._isMethodPayable(m),
         };
       });
```
rmeissner, line 81: Line 83 uses `fallback` to check for a `payableFallback` is this property still used? If so it should include `receive`, right?

package.json
```diff
@@ -35,7 +35,7 @@
     "@typescript-eslint/parser": "^4.19.0",
     "babel-eslint": "10.1.0",
     "dotenv-cli": "^4.0.0",
-    "eslint": "7.22.0",
+    "eslint": "^7.23.0",
```
rmeissner, line 38: Why is this required in this PR?

apps/wallet-connect/package.json
```diff
@@ -5,7 +5,7 @@
   "homepage": "./",
   "dependencies": {
     "@ethersproject/abi": "^5.0.13",
-    "@gnosis.pm/safe-react-components": "https://github.com/gnosis/safe-react-components.git#ff29c3c",
+    "@gnosis.pm/safe-react-components": "^0.5.0",
```
rmeissner, line 8: Why is this required in this PR?

rmeissner (changes requested):
We are doing some things in this PR which in my opinion would make sense to separate in different PRs (as "creating smaller PRs with better description" was a goal set in the retro):
- Update dependencies
- Refactor UI

The actual fix is currently 1 line of code and for this we have a PR with 1000+ line changes, is this really required?

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+      transactions.push({
+        description,
+        raw: { to: cleanTo, value: cleanValue, data },
+      });
+      setInputCache([]);
+      setTransactions(transactions);
+      setSelectedMethodIndex(0);
+      setValueInput('');
+    } catch (e) {
+      console.error(e);
```
rmeissner, line 156: Shouldn't we display an error here?

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+      }
+    }
+
+    try {
+      const cleanTo = web3.utils.toChecksumAddress(toInput);
+      const cleanValue = valueInput.length > 0 ? web3.utils.toWei(valueInput) : 0;
+      if (data.length === 0) {
+        data = '0x';
+      }
+      if (description.length === 0) {
```
rmeissner, line 144: Shouldn't we rather check if `data` is 0?

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+        if (i > 0) {
+          description += ', ';
+        }
+        const input = method.inputs[i];
+        description += (input.name || input.type) + ': ' + cleanValue;
+      }
+      description += ')';
+
+      try {
+        if (!['receive', 'fallback'].includes(method.name)) {
```
rmeissner, line 129: Should we skip the whole description/ method parsing (so have this check at the beginning of the if in line 113) and directly go the the next part?

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -0,0 +1,319 @@
...
+  // set when inputValue is visible
+  useEffect(() => {
+    const isVisible = async () => {
+      if (contract) {
+        const method = getContractMethod();
+        setIsValueInputVisible(method?.payable || false);
+      } else {
+        try {
+          const code = await services?.web3?.eth.getCode(toInput);
+          setIsValueInputVisible(code === '0x');
```
rmeissner, line 199: Why are we doing this? Couldn't we just allow it when we don't have an abi? We cannot really say if it is possible or not to send eth ... not sure if defaulting to "not allowing" it for contracts is the best choice.

rmeissner:
Sorry to be so harsh on this PR. I totally see the need to refactor this code, and pull the different logic parts into different components. Any we could even go further and move the transaction encoding / checking logic out of the components. 

I just feel that this does not belong in this PR that is for fixing the constructor issue. Even fixing the issue for sending ETH to an EOA should probably be handled in a separate PR.

matextrem:
@rmeissner you are right. This PR was merged with #issue-108 branch in order to avoid conflicts but it added a lot of changes from other PR made by @nicosampler. So I'm thinking about close this PR and create a new one with only the changes related to this bug.

rmeissner:
Ahhh yeah ... wasn't aware of the other PR (probably most of my comments from this PR should go there 😬). If you don't want to open a new PR you can also just rebase this PR and force push, but up to you ;). 

Another alternative: If you say this PR has a hard requirement on that other PR you make this PR a Draft PR that has the other PR branch as the base. As soon as the other PR is merged you can convert this PR to "Ready for review" and update the base (I think github even updates this PR base as soon as the other PR is merged).

PR #118, opened 2021-04-07
Transaction Builder - Change receive method to constructor

### LINKED ISSUE #109: [Transaction Builder] Constructor is rendered as receive

matextrem:
closes #109


apps/tx-builder/src/hooks/useServices/interfaceRepository.ts
```diff
@@ -76,14 +75,12 @@ class InterfaceRepository {
       .map((m: any) => {
         return {
           inputs: m.inputs || [],
-          name: m.name || (m.type === 'fallback' ? 'fallback' : 'receive'),
+          name: m.name || (m.type === 'fallback' ? 'fallback' : m.type === 'constructor' ? 'constructor' : 'receive'),
```
mikheevm, line 78: Could this include a little bit more context why it's needed? Comment or a function with a meaningful name that explains what it does

matextrem: You have the detailed info in #109, but basically the real method is `constructor` (looking at the ABI response from etherscan) instead of `receive`

mikheevm: Do you really expect someone searching half of GitHub when stumbling upon this unreadable piece of code? Yes, I do have the info, but someone other reading it won't.

mikheevm: Also, you cannot interact with a constructor but it's still shown
From #109 
>The real method would be the constructor (looking at the ABI response from etherscan), which is not interactable on deployed contracts.

matextrem: My understood of the problem was to show the method `constructor` with the parameter `_masterCopy`.  Could you @rmeissner clarify this to me?

rmeissner: We should not display the constructor as you cannot call it on a deployed contract. Constructirs are only available when deploying a new contract, but not during runtime as their code is not stored on chain.

We should still properly handle the receive method. E.g. if there is a payable receive or fallback function (or if we have no abi at all) we should display the eth field to generate a tx that only transfers eth. We should never show the constructor.

rmeissner: Also I think we should filter by function types we support. Currently we filter by the types we do not support (e.g. event) but if solidity adds a new type it will break the tx builder

nicosampler: > we should display the eth field to generate a tx that only transfers eth.

This was already done in my pr and is already merged.

matextrem: Updated!

mikheevm (commented):
LGTM but I would wait for an approval from @rmeissner as I miss some context

rmeissner (approved):

liliya-soroka:
I can not approve the fix as there are a few unclear cases for me :

- [ ] Contract address without ABI ( rinkeby : 0x976eed9d68b21544c61be7a2a8c80cde4aa5be11 )
Current result: 
image
Question: should not the receiver method be added for such contract address to send ETH to it ? - https://gnosisinc.slack.com/archives/CQB729LLF/p1619183150392000?thread_ts=1619182180.390600&cid=CQB729LLF 

- [ ] Mastercopy address ( 0x34CfAC646f301356fAa8B21e94227e3583Fe3F5F )
Current result: there is no receiver method only fallback . Is it correct ?

- [ ] ABI with payable method - https://gitlab.gnosisdev.com/wallets/gnosis-safe-qa/-/wikis/ABI/ABI-with-payable-function
Current result: No receiver method but as I understand from the slack discussion ( https://gnosisinc.slack.com/archives/CQB729LLF/p1619183150392000?thread_ts=1619182180.390600&cid=CQB729LLF) receiver method should be hidden only if no payable method in the ABI 

- [ ] do you have any contract address with recognised ABI to check receiver method fields?

rmeissner:
`0x34CfAC646f301356fAa8B21e94227e3583Fe3F5F` has a `payable fallback` method.

https://rinkeby.etherscan.io/address/0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552#code has a `receive` method and a `non-payable fallback` method.

https://rinkeby.etherscan.io/address/0x59AD6735bCd8152B84860Cb256dD9e96b85F69Da#code has no `payable` method add all.

tschubotz:
> For me it doesn't matter if a receive method is added or there is just the ETH input available. This is up for @tschubotz and @KristinaMayman to decide.

We never worked on the UX of this. So we don't have much insights here. The current solution with ETH value seems fine as is to me.

liliya-soroka:
Verified.
Thank you for your explanation.
@francovenica , I have added a few cases with the data from Richard to the Transaction builder test suite

PR #120, opened 2021-04-15
Drain Account app

### LINKED ISSUE #114: Drain Account
<Placeholder ticket, spec is just a draft>

Sometimes you just want to move on... This Safe App would allow to empty all assets from your Gnosis Safe account and send them somewhere else.

- Transfer all known ETH, ERC20, ERC721, etc. belonging to a Safe to a new address in one transaction
- Could be as easy as just an input field with a recipient address and a "Drain all assets" button or could also display a list of all assets where the user could select individual asset types to drain

katspaugh:
This PR adds a new Safe app to transfer all assets to another address.

I've also replaced Lerna with Yarn Workspaces because it was easier to get the tests run this way.

Screenshot 2021-04-15 at 15 55 18

Closes #114


katspaugh:

apps/drain-safe/.gitignore
```diff
@@ -0,0 +1,24 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
```
mikheevm, line 1: Do you need to ignore something or top-level .gitignore is enough?

apps/drain-safe/src/App.tsx
```diff
@@ -0,0 +1,163 @@
+import React, { useCallback, useState, useEffect } from 'react';
+import { Button, Loader, Title, TextField, Table, EthHashInfo } from '@gnosis.pm/safe-react-components';
+import { useSafeAppsSDK } from '@gnosis.pm/safe-apps-react-sdk';
+import Web3 from 'web3';
+import {AbiItem} from 'web3-utils';
```
mikheevm, line 5: Seems the formatting is off a bit

mikheevm: Ah the doesn't have a prettier pre-commit hook :(

apps/drain-safe/src/App.tsx
```diff
@@ -0,0 +1,163 @@
...
+import erc20 from './abis/erc20';
+import { fetchJson } from './utils';
+import useServices from './hooks/useServices';
+import Container from './Container';
+import Icon from './Icon';
+import Flex from './Flex';
+
+interface IAsset {
```
mikheevm, line 13: nitpick: we don't really use I* type of naming across the projects, would be good to stay consistent

apps/drain-safe/tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
mikheevm, line 3: I know this comes from a template that needs to be updated, but if you have time: can we bump this to es2015? most of the browsers support it nowadays

apps/drain-safe/src/Container.tsx
```diff
@@ -0,0 +1,13 @@
+import styled from 'styled-components';
+
+export default styled.form`
```
mikheevm, line 3: Do you know if this would be a violation of this rule? https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-anonymous-default-export.md

It can lead to some react-specific downsides: https://twitter.com/dan_abramov/status/1255229440860262400

I think in this specific case it would be shown as styled.form in the devtools, but not sure about the others 🤔

apps/drain-safe/src/Flex.tsx
```diff
@@ -0,0 +1,6 @@
+import styled from 'styled-components';
+
+export default styled.div`
```
mikheevm, line 3: same question as above

apps/drain-safe/src/hooks/useServices/index.ts
```diff
@@ -0,0 +1,40 @@
...
+import { rpcUrlGetterByNetwork } from '../../utils';
+
+export interface Services {
+  web3: Web3 | undefined;
+  interfaceRepo: InterfaceRepo | undefined;
+}
+
+export default function useServices(network: Networks): Services {
+  const [web3, setWeb3] = useState<Web3 | undefined>();
+  const [interfaceRepo, setInterfaceRepo] = useState<InterfaceRepository | undefined>();
```
mikheevm, line 16: `interfaceRepo` doesn't seem to be used anywhere

apps/drain-safe/src/App.tsx
```diff
@@ -0,0 +1,163 @@
...
+      } : {
+        // For other token types, generate a contract tx
+        to: web3!.utils.toChecksumAddress(item.tokenInfo.address),
+        value: '0',
+        data: encodeTxData(toAddress, item.balance)
+      };
+    });
+
+    const params = {
+      safeTxGas: 500000,
```
mikheevm, line 89: This should be hardcoded as this may vary across users: some may have 1 token, others may have 100 tokens. Gas usage for these txs will be different

apps/drain-safe/src/App.tsx
```diff
@@ -0,0 +1,163 @@
...
+      console.log({ safeTxHash });
+      setAssets([]);
+    } catch (e) {
+      console.error(e);
+    }
+    setSubmitting(false);
+  }, [sdk, assets, toAddress, web3, encodeTxData]);
+
+  const onToAddressChange = (e: React.ChangeEvent): void => {
+    const val = (e.target as HTMLInputElement).value;
```
mikheevm, line 103: ChangeEvent accepts a type argument for the element type
```suggestion
  const onToAddressChange = (e: React.ChangeEvent<HTMLInputElement>): void => {
    const val = e.target.value;
```

mikheevm (commented):
Good job!

apps/drain-safe/src/hooks/useServices/index.ts
katspaugh: It seems to be used in that file. Tbh, I just copied the hooks from TX Builder.

katspaugh:
Thanks Mikhail! I've fixed the issues and ran prettier on the app folder.

katspaugh:
What would be the best way to make sure the transaction succeeded and the assets are transferred?
I've tried polling the /balances endpoint but it seems to take very long to update.

mikheevm:
> What would be the best way to make sure the transaction succeeded and the assets are transferred?
> I've tried polling the /balances endpoint but it seems to take very long to update.

I think for now showing a message like "The transaction has been created, refresh the app when it is executed" is the most weighted solution between effort/quality 😃 Polling is also fine.

Multisig wallets UX is an ongoing research topic we have. For a regular wallet, you could just use promises to wait until the transaction is executed. For a multi-sig wallet, it's a little bit trickier because the transaction is created, but the execution can take some time (as other people could have to sign it)

katspaugh:
Thanks for the detailed answer, @mikheevm! Sounds reasonable.
I'll add the message you suggested.

apps/drain-safe/src/App.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+  return (
+    <Container>
+      <Flex>
+        <Title size="md">Drain Account</Title>
+        <div style={{ flex: 1 }} />
+        <EthHashInfo hash={safe.safeAddress} network={safe.network} textSize="lg" showIdenticon showCopyBtn />
+      </Flex>
+
+      {finishedTxHash && (
```
mikheevm, line 152: The transaction hash returned is a safe transaction hash, not the actual ethereum transaction hash, thus etherscan won't work. More info: https://docs.gnosis.io/safe/docs/contracts_tx_execution/#transaction-hash

Think of it as of internal transaction hash used by safe contracts to distinguish transactions. Also, when you approve a transaction in the interface, you actually approve the hash.

katspaugh: Oh, OK. I guess it doesn't make sense to display it then?

katspaugh:
@lukasschor any critique/suggestions from your side?

apps/drain-safe/src/App.tsx
mikheevm: Yes, etherscan link will be broken and confusing for the user. Only nerds will get it 😀

lukasschor:
Works very well and is definitely a valuable app! Few comments

1) The icon you used is open source / free to use, right?

2) The Safe address in the top-right is a bit redundant as it's already displayed in the sidebar. Is there a reason that it is again displayed there?

3) I would personally prefer a contained button at the bottom as it is the primary action and currently not really prominent. The style you used is usually applied to secondary actions in the app.

4) When I trigger a transaction and cancel the modal, the loading spinner will disappear and the input field is available again. Which is what I would expect. However, when I trigger a transaction and select one of the sub-transactions in the modal and only then close the modal, the loading spinner is spinning indefinitely. Not sure if this is an issue related to the app or that we do not communicate the closed modal via the SDK in case the user is in the detail view of one transaction (could be as we only launched this feature recently). Not that big of an issue though, as there is the cancel button.

CleanShot 2021-04-16 at 15 49 28@2x

Overall, nothing that I think necessarily needs to be addressed before merging this. Just some improvement suggestions.

katspaugh:
Thanks Lukas!

The icon is from here: https://www.flaticon.com/free-icon/withdraw_1682382 – I'ved added an attribution in the SVG file. ✅ 

Good point about the button, fixed! ✅

Regarding the infinite loader after closing the popup from the transaction details, I'm also not sure why this isn't rejecting the send promise. Might need to investigate in the SDK.

mikheevm:
> Regarding the infinite loader after closing the popup from the transaction details, I'm also not sure why this isn't rejecting the send promise. Might need to investigate in the SDK.

This sounds like a bug in the integration on safe-react side. Could you please create a ticket so we don't forget about this issue?

katspaugh:
Created https://github.com/gnosis/safe-react/issues/2178 👍

katspaugh:
Feedback from @rmeissner: need to add some tests for the logic to make the app production-ready.

apps/drain-safe/src/App.tsx
```diff
@@ -0,0 +1,192 @@
...
+      ) : (
+        <>
+          <TextField onChange={onToAddressChange} value={toAddress} label="Recipient" />
+
+          <Flex centered>
+            <Button
+              size="lg"
+              color="primary"
+              variant="contained"
+              onClick={submitTx}
```
nicosampler, line 180: does it make sense to add a confirmation modal?

apps/drain-safe/src/App.tsx
```diff
@@ -0,0 +1,192 @@
...
+      setAssets(data.items);
+    } catch (err) {
+      console.error("Couldn't load assets", err);
+    }
+  }, [safe]);
+
+  const submitTx = useCallback(async () => {
+    setSubmitting(true);
+    setFinished(false);
+
```
nicosampler, line 59: can we validate the address here? 
perhaps by calling `web3Utils.toChecksumAddress(toAddress)` and if it fails show an error in the input.

apps/drain-safe/src/App.tsx
mikheevm: no, it looks confusing because right after that safe-react will show a confirmation modal

nicosampler: that's correct. I forgot about that.

apps/drain-safe/src/App.tsx
katspaugh: Absolutely, good point 👌

mikheevm (approved):

apps/drain-safe/src/App.tsx
```diff
@@ -173,19 +193,13 @@ const App: React.FC = () => {
           <TextField onChange={onToAddressChange} value={toAddress} label="Recipient" />
```
nicosampler, line 161: nice, remember that `TextField` already handles errors. For this case isn't neccessary because you are handling a global error. 
https://safe-components.dev.gnosisdev.com/?path=/docs/inputs-textfield--simple-text-field#error

nicosampler (approved):

apps/drain-safe/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "drain-safe",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@gnosis.pm/safe-apps-react-sdk": "2.2.0",
+    "@gnosis.pm/safe-react-components": "^0.3.0",
```
dasanra, line 7: This makes me remember that maybe we should upgrade the template to safe-react-components v0.5.0 that it's the latest one. And maybe we should release v0.6.0 shortly, there are many improvements since then

apps/drain-safe/public/index.html
```diff
@@ -0,0 +1,40 @@
...
+    <!--
+      Notice the use of %PUBLIC_URL% in the tags above.
+      It will be replaced with the URL of the `public` folder during the build.
+      Only files inside the `public` folder can be referenced from the HTML.
+
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>Gnosis Safe App</title>
```
dasanra, line 24: Could we update the title here to something meaningful like `Drain Account app`?

apps/drain-safe/src/App.tsx
```diff
@@ -0,0 +1,174 @@
...
+  const onError = (userMsg: string, err: Error) => {
+    setError(`${userMsg}: ${err.message}`);
+    console.error(userMsg, err);
+  };
+
+  const fetchBalances = useCallback(async (): Promise<void> => {
+    // Fetch safe assets
+    try {
+      const data = await fetchSafeAssets(safe.safeAddress, safe.network);
+      console.log('Safe balances', data);
```
dasanra, line 29: Is this left on purpose?

apps/drain-safe/src/App.tsx
```diff
@@ -0,0 +1,174 @@
...
+    resetMessages();
+    setSubmitting(true);
+
+    const txs = assets.map((item) => tokenToTx(toAddress, item));
+
+    let safeTxHash = '';
+    try {
+      const data = await sdk.txs.send({ txs });
+      safeTxHash = data.safeTxHash;
+      console.log(safeTxHash);
```
dasanra, line 56: Is this left on purpose?

katspaugh:
Thanks @dasanra! Done ✅

apps/drain-safe/src/App.tsx
dasanra: For this you might need latest version of safe-react-components as I mentioned above

apps/drain-safe/package.json
```diff
@@ -0,0 +1,57 @@
...
+    "@gnosis.pm/safe-react-components": "^0.5.0",
+    "@material-ui/core": "^4.11.0",
+    "@testing-library/jest-dom": "^5.11.4",
+    "@testing-library/react": "^11.2.2",
+    "@testing-library/user-event": "^12.1.3",
+    "@types/jest": "^26.0.10",
+    "@types/node": "^14.6.2",
+    "@types/react": "^17.0.0",
+    "@types/react-dom": "^17.0.0",
+    "@types/styled-components": "^5.1.2",
```
dasanra, line 16: Would it make sense to organize this better splitting the ones that should be on devDependencies?

As now this is part of a mono-repo, should we remove those that are already in the common package.json and have no version differences?

apps/drain-safe/package.json
```diff
@@ -0,0 +1,57 @@
+{
+  "name": "drain-safe",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@gnosis.pm/safe-apps-react-sdk": "2.2.0",
+    "@gnosis.pm/safe-react-components": "^0.5.0",
+    "@material-ui/core": "^4.11.0",
```
dasanra, line 8: I think there is no direct use of this library in this app, so we could leave the version managing to safe-react-components?

dasanra (approved):
Just some small optimization comments. Overall very good work here :+1:

katspaugh:
Thanks @dasanra! I've removed the console logs and cleaned up the dependencies. 👍

apps/drain-safe/package.json
mikheevm: it's a peer dependency: https://github.com/gnosis/safe-react-components/blob/development/package.json#L85

mikheevm: so it has to be present

katspaugh: Restored 👌

apps/drain-safe/src/utils/gateway.ts
```diff
@@ -0,0 +1,29 @@
...
+interface Balance {
+  fiatTotal: string;
+  items: Asset[];
+}
+
+export const CURRENCY = 'USD';
+
+export async function fetchSafeAssets(safeAddress: string, safeNetwork: string): Promise<Balance> {
+  const network = safeNetwork.toLowerCase() === 'mainnet' ? 'mainnet' : 'rinkeby';
+  const url = `https://safe-client.${network}.gnosis.io/v1/safes/${safeAddress}/balances/${CURRENCY}/?trusted=false&exclude_spam=true`;
```
rmeissner, line 26: While this is indeed the easiest way, be aware that the client gateway is for the client, so we will deprecate endpoints that are not used by the mobile clients or react interface. We should find a better way for this in the future. I don't think it is blocking for now.

rmeissner (changes requested):

.github/workflows/deploy.pr.yml
```diff
@@ -48,7 +48,7 @@ jobs:
           yarn cache clean
 
       - name: Build apps
-        run: npx lerna run build
+        run: yarn build
```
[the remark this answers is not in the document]
katspaugh: Tests weren't being run properly as far as I could tell, so I "fixed" it by removing Lerna. Not a problem to restore if you think we'll need it in future.

apps/drain-safe/src/utils/gateway.ts
katspaugh: I'll check if I can use the transactions API directly.

.github/workflows/deployment.yml
```diff
@@ -51,7 +51,7 @@ jobs:
           yarn cache clean
 
       - name: Build apps
-        run: npx lerna run build
+        run: yarn build
```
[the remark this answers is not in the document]
katspaugh: Added ✅

katspaugh:
@rmeissner I've addressed your comments, thank you!

The test and lint jobs are failing with this error:
> error https://registry.yarnpkg.com/lodash/-/lodash-4.17.20.tgz: Extracting tar content of undefined failed, the file appears to be corrupt: "ENOENT: no such file or directory, stat '/usr/local/share/.cache/yarn/v6/npm-lodash-4.17.20-b44a9b6297bcb698f1c51a3545a2b3b368d59c52-integrity/node_modules/lodash/fp/compact.js'"

Looks like something on the Yarn repo side? I'll try to rebuild later today.

.github/workflows/deploy.pr.yml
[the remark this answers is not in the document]
rmeissner: I don'T really mind. @mikheevm set this up so he should know better, I personally prefer less dependencies, so if lerna is not required feel free to remove it :)

apps/drain-safe/src/utils/gateway.ts
```diff
@@ -0,0 +1,26 @@
...
+    decimals: number;
+    logoUri: string;
+  };
+}
+
+export const CURRENCY = 'USD';
+
+export async function fetchSafeAssets(safeAddress: string, safeNetwork: string): Promise<Asset[]> {
+  const network = safeNetwork.toLowerCase(); // mainnet, rinkeby etc
+  const url = `https://safe-transaction.${network}.gnosis.io/api/v1/safes/${safeAddress}/balances/${CURRENCY.toLowerCase()}/?exclude_spam=true`;
```
rmeissner, line 23: You have to be aware that on some networks this might not be available, so in that case the request will fail ;)

Minor nitpick: the file is called `gateway.ts` but we use don't use the gateway here :P

apps/drain-safe/src/utils/gateway.ts
rmeissner: What I meant is that theoretically you can use events to determine what tokens have been used with this address and fetch the balances for these tokens (that would be the version that does not rely on our services at all) :D

Again I don't think this is required for an initial version.

rmeissner (approved):

apps/drain-safe/src/utils/gateway.ts
katspaugh: Renamed 👍

liliya-soroka:
@katspaugh , should it go through QA? It was not tested and approved by QA

liliya-soroka:
Not approved by QA:

- [ ] Decimals are not taken into account for the token value:
Drain account app :

Actual values on balance page where decimals are taken into account:



- [ ] no default icon for the tokens without icon as on balance page :
Current result: __
Expected result: 

- [ ] What is the default sorting ?

katspaugh:
@liliya-soroka I thought we were supposed to test individual tickets ourselves and you only to do the regression.

Anyway, thanks for testing so quickly! Could you create a separate issue for everything together?
Except for the last point with the sorting arrow – it's an issue with the Table component from SRC.

apps/drain-safe/.env.sample
```diff
@@ -0,0 +1 @@
+PORT=3000
```
bh2smith, line 1: nanonit: newline at EOF.

katspaugh: Thanks Ben 😄  Fixed in #126.

liliya-soroka:
Verified

katspaugh:
Thanks @liliya-soroka!

PR #122, opened 2021-04-16
ETH Wrapper

### LINKED ISSUE #110: ETH Wrapper
- Wrap ETH -> WETH
- Unwrap WETH -> ETH
- Show balances of both ETH and WETH

jpalvarezl:
Closes #110 

Sorry if the PR is messy. Let me know what needs modifying to have it comply with the safe app submission process :) 

Thanks!


jpalvarezl:

mikheevm:
@jpalvarezl it would be great if you fix eslint errors 🙂

apps/eth-wrapper/README.md
```diff
@@ -0,0 +1,7 @@
+# Eth-Wrapper
+
+## Usage
+
+Simple safe app to wrap ETH or unwrap WETH.
```
rmeissner, line 5: Could you add how to build the app and run tests. E.g. 
- What npm version (or an `.nvmrc` file, see:  https://til.hashrocket.com/posts/nkdafd69bi-create-an-nvm-config-file)
- What package manager (e.g. yarn or npm)
- What command to run tests/ lint or similar

apps/eth-wrapper/README.md
```diff
@@ -0,0 +1,7 @@
...
+
+Add the Safe app at your own risk. This is hosted: [https://jpalvarezl.github.io/eth-wrapper/](https://jpalvarezl.github.io/eth-wrapper/)
```
rmeissner, line 7: I would make clear that this is not necessarily the same version as the one in this repo (unless you keep your repo in sync with this one)

apps/eth-wrapper/public/index.html
```diff
@@ -0,0 +1,43 @@
...
+    <!--
+      Notice the use of %PUBLIC_URL% in the tags above.
+      It will be replaced with the URL of the `public` folder during the build.
+      Only files inside the `public` folder can be referenced from the HTML.
+
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>ethWrapper Safe App</title>
```
rmeissner, line 27: Shouldn't that be a capital `E`?

apps/eth-wrapper/public/logo.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 841.9 595.3">
```
rmeissner, line 1: We should remove the React icons and point everything that still used these to the eth wrapper icon.

apps/eth-wrapper/src/assets/svg/weth.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="2000" height="2000" viewBox="0 0 2000 2000"><rect x="184.16" width="1815.84" height="1815.84" rx="267.39" fill="#939393"/><rect y="184.16" width="1815.84" height="1815.84" rx="267.39" fill="#303030"/><path d="M907.79,367.78l-9.71,33v957.83l9.71,9.69,444.59-262.79Z" fill="#9b9b9b"/><path d="m907.79 367.78l-444.59 737.74 444.59 262.81" fill="#fff"/><path d="M907.79,1452.52l-5.48,6.68v341.25l5.48,15.93,444.85-626.55Z" fill="#9b9b9b"/><path d="m907.79 1816.38l0-363.87-444.59-262.68" fill="#fff"/><path d="M907.79,1368.33l444.58-262.81L907.79,903.43v464.9" fill="#666"/><path d="m463.2 1105.52l444.58 262.81 0-464.9" fill="#9b9b9b"/></svg>
```
rmeissner, line 1: This is not really the official WETH icon, right? (If it is than the app icon should also be this one)

apps/eth-wrapper/src/components/EthBalance.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+    async function fetchBalance() {
+        const balanceEth = await sdk.eth.getBalance([safe.safeAddress]);
+        setBalance(ethers.utils.formatEther(balanceEth));
+    }
+
+    useEffect(() => {
+        fetchBalance();
+        console.log("Updating ETH balance");
+    });
```
rmeissner, line 18: shouldn't this have a dependency array?

apps/eth-wrapper/src/components/EthBalance.tsx
```diff
@@ -0,0 +1,24 @@
+import React, { useEffect, useState } from 'react';
+import { ethers } from 'ethers';
+import { useSafeAppsSDK } from '@gnosis.pm/safe-apps-react-sdk';
+import { Text } from '@gnosis.pm/safe-react-components';
+
+const EthBalance: React.FC = () => {
+    const [balance, setBalance] = useState("loading...");
+    const { sdk, safe } = useSafeAppsSDK();
+
+    async function fetchBalance() {
```
rmeissner, line 10: I am not sure when useCallback is better, but we could probably also declare it inside the useEffect.

But currently if this fails the whole app crashes, right?

apps/eth-wrapper/src/components/WethBalance.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+    async function fetchBalance() {
+        const balanceWeth = await weth.balanceOf(safe.safeAddress);
+        setBalance(ethers.utils.formatEther(balanceWeth));
+    }
+
+    useEffect(() => {
+        fetchBalance();
+        console.log("Updating WETH balance");
+    });
```
rmeissner, line 23: Same comments as for the EthBalance component

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+const Wrapper: React.FC<WrapperProps> = (props: WrapperProps) => {
+    const { sdk, safe } = useSafeAppsSDK();
+
+    const [amountToWrap, setAmountToWrap] = useState("");
+    const [errorMessage, setErrorMessage] = useState("");
+    const [availableBalance, setAvailableBalance] = useState(0.0);
+    const [isError, setIsError] = useState(false);
+    const [safeTxHash, setSafeTxHash] = useState("");
+
+    const provider = useMemo(() => new SafeAppsSdkProvider(safe, sdk), [safe, sdk]);
```
rmeissner, line 24: We create this provider in multiple places wouldn't it make more sense to have it somewhere more global?

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+    const weth = useMemo(() => new ethers.Contract(getWethAddress(safe.network.toLowerCase()), Erc20, provider), [provider, safe]);
```
rmeissner, line 25: It might be better to create the contract first without a provider and then just connect to it. This will minimize the creation overhead (same for the WethBalance component).

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+                setSafeTxHash(safeTx.safeTxHash);
+                console.log(safeTx.safeTxHash);
+            } catch (e) {
+                console.error(e)
+            }
+        } else {
+            try {
+                const parsedAmount = ethers.utils.parseEther(amountToWrap);
+                const withdraw = new ethers.utils.Interface(WETHwithdraw_function);
+                const safeTx = await sdk.txs.send({
```
rmeissner, line 51: We could lift the code with `send` and following out of the if/else to avoid code duplication.

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+                    }]
+                })
+                setSafeTxHash(safeTx.safeTxHash);
+                console.log(safeTx.safeTxHash);
+            } catch (e) {
+                console.error(e)
+            }
+        } else {
+            try {
+                const parsedAmount = ethers.utils.parseEther(amountToWrap);
```
rmeissner, line 49: This is quite confusing as it is actually the `amountToUnwrap`

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+                const withdraw = new ethers.utils.Interface(WETHwithdraw_function);
```
rmeissner, line 50: Why not reuse the WETH abi here?

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+            } catch (e) {
+                console.error(e);
+            }
+        }
+    }, [sdk, amountToWrap, isError, props.wrap, safe])
+
+    const validateAmout = useCallback((newValue: string) => {
+        console.log("available balance:", availableBalance);
+        console.log("new value: ", newValue)
+        if (isNaN(Number(newValue))) {
```
rmeissner, line 69: why not try `parseEther`? wouldn'T that be a more reliable check if we can use it later?

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+            setIsError(true);
+            setErrorMessage("Not a number");
+        }
+        else if (Number.parseFloat(newValue) > availableBalance) {
```
rmeissner, line 73: Here I would use BigNumber comparision instead of converting both to a float and comparing them

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+            // update available balance
+            var newValue = "0";
+            if (props.wrap) {
+                const balanceEth = await sdk.eth.getBalance([safe.safeAddress]);
+                newValue = ethers.utils.formatEther(balanceEth);
+            } else {
+                const balanceWeth = await weth.balanceOf(safe.safeAddress);
+                newValue = ethers.utils.formatEther(balanceWeth);
+            }
+            setAvailableBalance(Number.parseFloat(newValue));
```
rmeissner, line 99: I would recommend `parseEther` here and render that with `toString`

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+        runEffect();
+    }, [safe, sdk, weth, props.wrap]);
+
+    return (
+        <Grid container spacing={3}>
+            <Snackbar
+                open={safeTxHash.length !== 0}
+                autoHideDuration={3000}
+                onClose={() => setSafeTxHash("")}
+                message="You transaction has been submitted"
```
rmeissner, line 110: Your

apps/eth-wrapper/src/components/Wrapper.tsx
```diff
@@ -0,0 +1,127 @@
...
+            />
+            <Grid item xs={8}>
+                <TextField
+                    value={amountToWrap}
+                    label={props.wrap ? "ETH amount" : "WETH amount"}
+                    meta={{ error: errorMessage }}
+                    onChange={e => validateAmout(e.target.value)} />
+            </Grid>
+            <Grid item xs={8}>
+                <Button size="md" color="primary" variant="contained" onClick={() => wrapEth()}>
```
rmeissner, line 120: Minor comment, but I see it in multiple parts: It is always called `wrapEth` which does not describe what it is actually doing.

apps/eth-wrapper/src/utils/Erc20Constants.ts
```diff
@@ -0,0 +1,11 @@
+export const Erc20 = [
+    "function approve(address _spender, uint256 _value) public returns (bool success)",
+    "function allowance(address _owner, address _spender) public view returns (uint256 remaining)",
+    "function balanceOf(address _owner) public view returns (uint256 balance)",
+    "event Approval(address indexed _owner, address indexed _spender, uint256 _value)"
+];
+
+export function getWethAddress(network: string): string {
+    return network === "mainnet" ? "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2" :
```
rmeissner, line 9: I would use a switch case or mapping and make it so that it fails if it is an unknown network. Currently on all "non-mainnet" networks we would use the rinkeby address ... which would cause loss of funds on networks like xdai.

apps/eth-wrapper/.env.sample
```diff
@@ -0,0 +1,3 @@
+PORT=3000
+HTTPS=true
+REACT_APP_INFURA_ID=
```
rmeissner, line 3: Is this used?

apps/eth-wrapper/.env.sample
```diff
@@ -0,0 +1,3 @@
+PORT=3000
+HTTPS=true
```
rmeissner, line 2: I was made aware that this is actually not required

rmeissner (changes requested):
I general nice PR but imo some cleanup is missing:

- Properly handle unsupported networks
- Remove unused assets
- Decide on either `npm` or `yarn` currently both package locks are included
- Add tests for some of the logic method

apps/eth-wrapper/package.json
```diff
@@ -0,0 +1,58 @@
...
+    "@testing-library/react": "^11.2.2",
+    "@testing-library/user-event": "^12.1.3",
+    "@types/jest": "^26.0.10",
+    "@types/node": "^14.6.2",
+    "@types/react": "^17.0.0",
+    "@types/react-dom": "^17.0.0",
+    "@types/styled-components": "^5.1.2",
+    "ethers": "^5.0.29",
+    "react": "^17.0.1",
+    "react-app-rewired": "^2.1.6",
```
rmeissner, line 21: This should be a dev dependency and actually it is not required in this setup as this project is part of a monorepo. See the other projects for how it should look like.

apps/eth-wrapper/src/assets/svg/weth.svg
jpalvarezl: ah yeah, I don't remember exactly, where I got that from, but I think I ended up using the one returned for weth from our services `TokenInfo`. I will get rid of it :)

apps/eth-wrapper/.env.sample
jpalvarezl: nope, will remove

apps/eth-wrapper/package.json
jpalvarezl: dependency removed. I modified scripts to be like the other apps.

apps/eth-wrapper/README.md
jpalvarezl: I will remove this, as I suppose having this deployed in my personal github pages is not ideal

apps/eth-wrapper/src/components/EthBalance.tsx
jpalvarezl: This is just the label with the balance. You would have to navigate away from the `eth-wrapper` app , to run a transaction and affect your balance. Unless, it is expected that I keep this value updated, in which case the dependency would be `sdk`?

apps/eth-wrapper/src/components/EthBalance.tsx
jpalvarezl: I would have to write a test for that, as I have no idea. I think that `useEffect` doesn't play nicely with inlined `async` functions. That's why it's extracted. I don't know if there is a more appropriate pattern for this.

apps/eth-wrapper/src/components/Wrapper.tsx
jpalvarezl: I don't know how to make this available globally. I assumed that `safe` and `sdk` were a sort of shared state that didn't have an initialization cost. I will look into that.

apps/eth-wrapper/src/components/Wrapper.tsx
jpalvarezl: This was addressed, I believe.

bh2smith:
Hey @jpalvarezl - I saw that you were making an ETH wrapper app and thought you might want to resuse some of the components from the one @W3stside and I had put together in the past. 

https://github.com/bh2smith/safe-wraptor

I haven't looked too closely at the style you went with, but the one I linked uses a bunch of safe-react-app components already. Cheers!

apps/eth-wrapper/src/utils/Erc20Constants.ts
```diff
@@ -0,0 +1,17 @@
...
+];
+
+export function getWethAddress(network: string): string {
+    switch (network.toLocaleLowerCase()) {
+        case "mainnet": 
+            return "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2";
+        case "rinkeby":
+            return "0xc778417E063141139Fce010982780140Aa0cD5Ab";
+        default:
+            throw new Error("Unsupported network");
```
bh2smith, line 15: You should also be able to support xdai and wxdai here on the xdai network.

bh2smith: I believe this is the address:

`0xe91D153E0b41518A2Ce8Dd3D7944Fa863463a97d`

apps/eth-wrapper/src/utils/Erc20Constants.ts
```diff
@@ -0,0 +1,17 @@
+export const Erc20 = [
+    "function approve(address _spender, uint256 _value) public returns (bool success)",
+    "function allowance(address _owner, address _spender) public view returns (uint256 remaining)",
+    "function balanceOf(address _owner) public view returns (uint256 balance)",
+    "event Approval(address indexed _owner, address indexed _spender, uint256 _value)"
+];
```
bh2smith, line 6: Should be able to import the erc20 interface like this:

```
import IERC20 from "@openzeppelin/contracts/build/contracts/IERC20.json";
```

apps/eth-wrapper/src/utils/WETHConstants.ts
```diff
@@ -0,0 +1,282 @@
+export const WETHwithdraw_function = ["function withdraw(uint wad) public"];
+export const WETH_ABI = `
```
bh2smith, line 2: Or maybe this is the abi interface:

import IERC20 from "@openzeppelin/contracts/build/contracts/IERC20.json";

apps/eth-wrapper/src/utils/Erc20Constants.ts
mikheevm: I like Jose's way more because it won't include the JSON to the bundle and means less code shipped to the users. Maybe there are some optimisations for JSON so I suggest to test it

PR #126, opened 2021-05-03
Fix balances display in Drain Safe

### LINKED ISSUE #114: Drain Account

katspaugh:
Closes #114.

Fixes some of the issues found by @liliya-soroka in #120:

- [x] Decimals are not taken into account for the token value:
Drain account app :

Actual values on balance page where decimals are taken into account:

- [x] no default icon for the tokens without icon as on balance page :
Current result: __
Expected result:


katspaugh:
NB: for missing icons, I'm just not displaying any image.

apps/drain-safe/src/components/Icon.tsx
```diff
@@ -12,9 +12,11 @@ const IconImg = styled.img`
   width: auto;
 `;
 
-function Icon(props: Props): JSX.Element {
+function Icon(props: Props): JSX.Element | null {
+  const [noIcon, setNoIcon] = useState<boolean>(false);
   const { logoUri, symbol } = props;
-  return <IconImg src={logoUri} alt={symbol} />;
+
+  return noIcon ? null : <IconImg src={logoUri} alt={symbol} onError={() => setNoIcon(true)} />;
```
dasanra, line 19: Would it make sense to use the 'assets' icon from safe-react-components inside a greyed circle, similar as we do in the assets tab?
https://components.gnosis-safe.io/?path=/docs/data-display-icon--icons

image

image

dasanra (commented):
I can still see the arrow for sorting when hovering table titles. What do you think about removing it as is not possible to do frontend sorting?

katspaugh:
> I can still see the arrow for sorting when hovering table titles. What do you think about removing it as is not possible to do frontend sorting?

The arrow cannot be removed as of now. I'll fix this in the SRC repo. See https://github.com/gnosis/safe-react-components/issues/128

apps/drain-safe/src/components/Icon.tsx
katspaugh: The assets icon is used in the sidebar menu so I think it might be confusing.
I've added a question mark icon as a fallback (like Uniswap).

Screenshot 2021-05-05 at 09 24 26

dasanra (approved):

liliya-soroka:
Verified
nice app

PR #132, opened 2021-06-04
Sync apps with master branch

dasanra:
the original issue - https://github.com/gnosis/safe-react-apps/issues/133


liliya-soroka:
@dasanra , do I understand correctly that this is the ticket for the Transaction builder full regression?
Do we need to run tests for other apps in the list?

mikheevm (approved):

Issue #131, opened 2021-06-04
[WalletConnect] Add Abitrum support


tschubotz:
# Context

Currently, the WalletConnect Safe app doesn't support dapps on Arbitrum. 

# Task

Please add support for Abitrum. See #127 for how this was done for BSC.

# Misc

- Some technical infos on Arbitrum [here](https://docs.google.com/document/d/1Sr0TP7WTjpm_K_MTklLrZM7gui6C44kL-p26keHt29g/edit).
- #128 is related also

Issue #128, opened 2021-05-10
[WalletConnect] Add Polygon support


tschubotz:
# Context

Currently, the WalletConnect Safe app doesn't support dapps on Polygon such as https://app.aave.com/.

# Task

Please add support for Polygon. See #127 for how this was done for BSC.

# Misc

Some technical infos on Polygon [here](https://docs.google.com/document/d/1PyOUWxDCZn9-ulHsV_bT9zRMTiGe9TnERmVLjIlK_OY/edit#).

katspaugh:
Estimation: 5

tschubotz:
This might be solvable easily with an update of the Safe apps SDK
https://github.com/gnosis/safe-apps-sdk/pull/162
https://github.com/gnosis/safe-react/pull/2288

PR #136, opened 2021-06-28
Feature: Support any network for wallet-connect

### LINKED ISSUE #127: [WalletConnect] Add BSC support
Currently it is not possible to use the Wallet Connect app on BSC as it is not using the correct chainid.

image

Possible solution:
- Adjust https://github.com/gnosis/safe-react-apps/blob/development/apps/wallet-connect/src/utils/networks.ts to include BSC
- Fallback to loading chainid via an rpc call through the interface.

mikheevm:
This PR adjusts walletconnect app to support any network by using chainId passed from the SDK

Closes #127 #128 #131


apps/wallet-connect/src/hooks/useWalletConnect.tsx
```diff
@@ -41,7 +38,7 @@ const useWalletConnect = () => {
 
         wcConnector.approveSession({
           accounts: [safe.safeAddress],
-          chainId: chainIdByNetwork[network],
+          chainId: safe.chainId,
```
katspaugh, line 41: That was easy 😄

katspaugh (approved):

dasanra (approved):

apps/wallet-connect/src/hooks/useWalletConnect.tsx
kenyl899: ### It's really nice

francovenica:
Tested it in binance, in the PR 2460. I added the Wallet connect app and connected with the BSC block explorer. I tried a tx that end up failing, but still it shows that you can use the wallet connect app.
image

PR #137, opened 2021-06-28
[TX-Builder] Always show to field

dasanra:
We are having some limitations on the way that transaction builder can be used. We decided to not show the to field when address was a valid address for interacting, but many people is loading de ABI using MasterCopy address or similar methods.

Restore to input field, even we show a duplicated value to allow people to still use the transaction builder this way

### DIFF
--- a/apps/tx-builder/package.json
+++ b/apps/tx-builder/package.json
@@ -1,6 +1,6 @@
 {
   "name": "tx-builder",
-  "version": "1.1.0",
+  "version": "1.2.0",
   "private": true,
   "homepage": "./",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


francovenica:
Tested comparing the deafault Tx builder with the one in the pr #137 with the master copy 1.1.1 and a MM account address
0x34CfAC646f301356fAa8B21e94227e3583Fe3F5F
0x61a0c717d18232711bC788F19C9Cd56a43cc8872

For the old one the "to" field doesn't show up, and for the fixed version, it DOES show up, which is what we were looking for
06-29-2021_x(2663)

katspaugh (approved):
👍

katspaugh:
@dasanra please merge and release

liliya-soroka:
For the history:
"To " field will be available as for the contract addresses, as for EOA

PR #138, opened 2021-06-29
Upgrade dependencies

dasanra:
Upgrade dev dependencies and fix eslint


.eslintrc.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   parser: '@typescript-eslint/parser',
-  extends: ['plugin:@typescript-eslint/recommended', 'prettier/@typescript-eslint', 'plugin:prettier/recommended'],
+  extends: ['plugin:@typescript-eslint/recommended', 'prettier', 'plugin:prettier/recommended'],
```
mikheevm, line 3:
```suggestion
  extends: ['plugin:@typescript-eslint/recommended', 'prettier'],
```
`prettier` is enough: https://github.com/prettier/eslint-plugin-prettier

package.json
```diff
@@ -2,11 +2,13 @@
   "name": "safe-apps",
   "private": true,
   "scripts": {
+    "build": "yarn workspaces run build",
+    "lint:check": "eslint './apps/**/*.{js,jsx,ts,tsx}'",
+    "lint:fix": "yarn lint:check --fix",
```
katspaugh, line 7: It would be nice to also workspace-ify the listing, so that we can the apps aren't affecting each other.

katspaugh (approved):

package.json
mikheevm: What do you mean by affecting each other? It doesn't make sense to use different style guides in a monorepo, does it? I hope I got it right

katspaugh: I meant when you run the fix/prettier, it would be nice to not fix/prettify the apps you haven't touched. But yarn workspaces probably won't help here anyway.
Please disregard my comment.

PR #142, opened 2021-07-05
[Tx-Builder] Fix negative input number

### LINKED ISSUE #139: Tx Builder: Not possible to use negative input for ints
CleanShot 2021-06-30 at 10 20 04@2x

A user reported that they are not able to use negative numbers for ints input values in the transaction buiilder.

To reproduce, use `0x034a6626eec82e14ac5b0130f04f85a2ea158f9c` as the contract (ABI source) and selected the `createPool` function.

dasanra:
Fixes #139 parsing correctly negative number on inputs

From web3 > 1.2.5 there is a bug that parse negative numbers incorrectly adding leading 0's before the minus sign when the value is passed as a string.

The encodeFunctionCall expects that the input values are still passed as string, so to avoid this bug without rolling back to a really old version of web3 we can add the leading 0's for numeric values and that way still send the value as a string

### DIFF
--- a/apps/tx-builder/src/components/Builder.tsx
+++ b/apps/tx-builder/src/components/Builder.tsx
@@ -59,7 +59,7 @@ const paramTypeNumber = new RegExp(/^(u?int)([0-9]*)$/);
 const parseInputValue = (input: any, value: string): string => {
   // If there is a match with this regular expression we get an array value like the following
   // ex: ['uint16', 'uint', '16']. If no match, null is returned
-  const isNumberInput = input.type.match(paramTypeNumber)
+  const isNumberInput = paramTypeNumber.test(input.type)
 
   if (value.charAt(0) === '[') {
     return JSON.parse(value.replace(/"/g, '"'));
@@ -67,11 +67,11 @@ const parseInputValue = (input: any, value: string): string => {
     // From web3 1.2.5 negative string numbers aren't correctly padded with leading 0's.
     // To fix that we pad the numeric values here as the encode function is expecting a string
     // more info here https://github.com/ChainSafe/web3.js/issues/3772
-    const bitWidth = isNumberInput[2]
+    const bitWidth = input.type.match(paramTypeNumber)[2]
     return toBN(value).toString(10, bitWidth);
-  } else {
-    return value;
   }
+  
+  return value;
 }
 
 type Props = {
--- a/apps/tx-builder/src/components/Builder.tsx
+++ b/apps/tx-builder/src/components/Builder.tsx
@@ -133,15 +133,15 @@ export const Builder = ({ contract, to }: Props): ReactElement | null => {
         const parsedInputs: string[] = [];
         const inputDescription: string[] = [];
 
-        method.inputs.forEach((input, index) => {
-          const cleanValue = inputCache[index] || '';
-          parsedInputs[index] = parseInputValue(input, cleanValue);
-          inputDescription[index] = `${(input.name || input.type)}: ${cleanValue}`;
-        })
-
-        description = `${method.name} (${inputDescription.join(', ')})`;
-
         try {
+          method.inputs.forEach((input, index) => {
+            const cleanValue = inputCache[index] || '';
+            parsedInputs[index] = parseInputValue(input, cleanValue);
+            inputDescription[index] = `${(input.name || input.type)}: ${cleanValue}`;
+          })
+
+          description = `${method.name} (${inputDescription.join(', ')})`;
+
           data = web3.eth.abi.encodeFunctionCall(method as AbiItem, parsedInputs);
         } catch (error) {
           setAddTxError(error.message);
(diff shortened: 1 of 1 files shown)


apps/tx-builder/src/components/Builder.tsx
```diff
@@ -57,6 +52,33 @@ const getInputHelper = (input: any) => {
...
 };
 
+// Same regex used for web3@1.3.6
+const paramTypeNumber = new RegExp(/^(u?int)([0-9]*)$/);
+
+// This function is used to apply some parsing to some value types
+const parseInputValue = (input: any, value: string): string => {
+  // If there is a match with this regular expression we get an array value like the following
+  // ex: ['uint16', 'uint', '16']. If no match, null is returned
+  const isNumberInput = input.type.match(paramTypeNumber)
```
katspaugh, line 62: It's better to use `test` because it returns a boolean.
```
const isNumberInput = paramTypeNumber.test(input.type)
```

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -57,6 +52,33 @@ const getInputHelper = (input: any) => {
...
+
+  if (value.charAt(0) === '[') {
+    return JSON.parse(value.replace(/"/g, '"'));
```
katspaugh, line 65: Parse can throw, is that OK in this case?

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -57,6 +52,33 @@ const getInputHelper = (input: any) => {
...
+  } else if (isNumberInput) {
+    // From web3 1.2.5 negative string numbers aren't correctly padded with leading 0's.
+    // To fix that we pad the numeric values here as the encode function is expecting a string
+    // more info here https://github.com/ChainSafe/web3.js/issues/3772
+    const bitWidth = isNumberInput[2]
+    return toBN(value).toString(10, bitWidth);
+  } else {
```
katspaugh, line 72: I would return this else and just do the default return.

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -108,21 +130,19 @@ export const Builder = ({ contract, to }: Props): ReactElement | null => {
...
+        const parsedInputs: string[] = [];
+        const inputDescription: string[] = [];
+
+        method.inputs.forEach((input, index) => {
+          const cleanValue = inputCache[index] || '';
+          parsedInputs[index] = parseInputValue(input, cleanValue);
+          inputDescription[index] = (input.name || input.type) + ': ' + cleanValue;
+        })
+
+        description = method.name + ' (' + inputDescription.join(', ') + ')';
```
katspaugh, line 142: It would be nice to use template strings instead of concatenation.

dasanra: We synced there, was adding that update in my last check :laughing:

apps/tx-builder/src/components/Builder.tsx
dasanra: Ok in any case I will need to extract the last numbers to get the bitWidth

katspaugh: Oh, didn't notice that. I would rename the variable into `numberMatch` or something.

apps/tx-builder/src/components/Builder.tsx
dasanra: To be honest I just left this code as it was. For a cleaner UX we should catch this error and return feedback to the user. If we catch this error and don't handle it on a specific way we could make the user to add a transaction with an invalid value. 

So to handle this properly would be out of scope for this ticket

katspaugh: Might be worth catching it in the function that calls `parseInputValue` maybe?

dasanra: This is the output we would get in that case. Maybe is more helpful than nothing, but still not really clear:

image

I added this ticket just in case
https://github.com/gnosis/safe-react-apps/issues/143

apps/tx-builder/src/components/Builder.tsx
dasanra: I did the `test` for checking that is a number and the `match` when I extract the `bitWidth` to be more explicit. Code readability is improved

apps/tx-builder/src/components/Builder.tsx
dasanra: Done :+1:

katspaugh (approved):

francovenica:
I added the Tx-builder given in this ticket to prod mainnet so I could try the same scenario using the contract and method described in the issue

Contract: 0x034a6626eec82e14ac5b0130f04f85a2ea158f9c 
Test Values:
CreatePool (method)
Address1: 0x6B175474E89094C44Da98b954EedeAC495271d0F
Address2: 0x6B175474E89094C44Da98b954EedeAC495271d0F
NumValue1: 100
NumValue2: 100
NumValue3: -10000
NumValue4: -1000
String: "1"

In the new tx builder the negative values are not treated as invalid values

Gif:
I show first the error, with the txbuilder that is in prod by default. The other tab shows the tx already created and ready to be siged, using the tx-builder app provided in this ticket, also in mainnet eth prod.
07-06-2021_x(2706)

Looks good to me

PR #150, opened 2021-08-04
[WC] Upgrade the client to the latest version

katspaugh:
WalletConnect is asking dapps to update their integration: https://twitter.com/WalletConnect/status/1422169112751529989

Screenshot 2021-08-04 at 10 46 21

It looks like they are working on v2 (there's a beta published on NPM). Maybe that would be even better but for now, let's just upgrade to the latest package.

### DIFF
--- a/apps/wallet-connect/package.json
+++ b/apps/wallet-connect/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "@ethersproject/abi": "^5.0.13",
     "@gnosis.pm/safe-react-components": "^0.5.0",
-    "@walletconnect/client": "^1.3.6",
+    "@walletconnect/client": "^1.5.2",
     "date-fns": "^2.17.0",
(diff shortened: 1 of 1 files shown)


mikheevm (approved):
good catch!

katspaugh:
FWIW, I've tested it on mainnet with SushiSwap. Seems to work a bit better than before (confirmation popup appearing faster).

odd-amphora:
Any idea when this will be merged?

francovenica:
I was able to use the App in rinkeby and mainnet:
Tx 1163 in this safe https://rinkeby.gnosis-safe.io/app/#/safes/0x9913B9180C20C6b0F21B6480c84422F6ebc4B808/transactions
and Tx 201 in this safe https://gnosis-safe.io/app/#/safes/0x8675B754342754A30A2AeF474D114d8460bca19b/transactions

Looks good to me

PR #154, opened 2021-08-11
Fix CI deployment

### LINKED ISSUE #149: Fix CI deployment for Staging and Prod
Currently we are only deploying test links for development, staging and prod services weren't fully migrated from Travis CI

We want to still keep alive a copy of centralized apps server so we have to adapt staging and prod hooks to behave as expected

### Necessary changes:
- Deploy staging versions into s3://STAGING_BUCKET_NAME/current
- Deploy tagged builds into s3://STAGING_BUCKET_NAME /releases
- Call hook when building on tag PROD_DEPLOYMENT_HOOK_URL

This is the same approach followed by safe-react repo. We may need to add [this script](https://github.com/gnosis/safe-react/blob/dev/scripts/github/prepare_production_deployment.sh)

Related
https://gitlab.gnosisdev.com/devops/aws/aws-infra-staging/gnosis-staging/-/issues/258

dasanra:
Resolves #149 

It adds the missing configuration to complete staging deployments and also the missing script and hooks to allow devops to have the latest tagged build available to deploy on their deployment service

In this PR I also updated branch `master` to `main` with the necessary changes in scripts

I unified all deployments into one Github action, as common steps are easier to keep up to date this way
I added a cancel-previous-build step to avoid overflowing CI when we add commits before previous actions finishes


katspaugh (approved):
Looks good!

Can we test it before merging? By using this branch's name instead of main maybe?

.github/workflows/deploy.pr.yml
```diff
@@ -1,79 +0,0 @@
-name: Deploy PR
```
mikheevm, line 1: Do you feel it's easier to maintain all deployments in a single deployment workflow?

mikheevm: I thought it's harder and this is why I split them. Even if there's some duplication I think it's easier to debug

dasanra: For me was adding more work load to keep both files synced than executing only the part that I want to test, as this can easily be done when something is not working commenting out the steps you don't need to check.

Also if the "if" logic is working as expected we should be able to know when each flow should run without issues.

Now it's following the same pattern as we have on `safe-react` so should be easier to keep all of them up to date

dasanra:
> Looks good!
> 
> Can we test it before merging? By using this branch's name instead of main maybe?

Already tested them while working changing the names to execute in the current branch and creating a testing tag that I validated with Devops

Issue #156, opened 2021-08-18
[QA test] Prepare a bug-ridden Safe App for QA hiring


katspaugh:
* Pie chart of assets allocation
* Legend for the chart
* Fiat currency dropdown
* Threshold input to hide small assets
* If time permits – a summary of last transactions

cc @liliya-soroka

PR #157, opened 2021-08-23
Safe Dashboard app

katspaugh:
See #156 – this app is for QA hiring. The low quality of the code is a feature. 😉 

Screenshot 2021-08-23 at 15 10 13


mikheevm:
Should this live on a separate branch? Not sure if it's reasonable to keep it along with other apps.

katspaugh:
> Should this live on a separate branch? Not sure if it's reasonable to keep it along with other apps.

Yeah, sure, I just created this PR to deploy it. We don't have to merge it.

PR #180, opened 2021-11-05
[TX Builder] Transactions are initialised when entering a contract with a wrong address (Abi not retrieved properly) 

### LINKED ISSUE #141: Transaction builder forgets added transactions when adding new ABI or address
Hi there, 

This is the issue is occurring on:

app url:
https://gnosis-safe.io/app/#/safes/0x0013CEFc104698Bb40d574E2aE6C822d5D52cff3/apps?appUrl=https://cloudflare-ipfs.com/ipfs/QmTJMCBZHX56z36aTUaL2QZ3rKTkzaq1CG3zhPSKb1Su3a

Brave version:
Release Notes V1.26.74 (June 30, 2021)

Description:

After the first transaction has been added, the front end reflects this by showing (1) in the send transactions button. Then if a new ABI or Address is interested in to the "Enter Address or ABI" field, the UI reverts to "Send transactions" without the 1. 

Please find the video attached. 

https://user-images.githubusercontent.com/7915365/124475485-0f88dc00-dda2-11eb-8aa0-1cd1a6cc7e73.mov

yagopv:
## What it solves
Resolves #141 

## How this PR fixes it
The problem was the component Builder being destroyed when the contract address entered is wrong. We moved the transaction management to the parent so they are kept

## How to test it
1) Open the transaction builder inside a Safe web and add a correct contract address for example 0x39aa39c021dfbae8fac545936693ac917d5e7563 (`mainnet`)
2) Add transactions using the `redeemUnderlying` method for example like in the video in #141 
3) Now enter a wrong address (For example deleting some characters) and then a correct one again
4) Transactions should be kept and when trying to submit the data associated should be ok

### DIFF
--- a/apps/tx-builder/src/components/Dashboard.tsx
+++ b/apps/tx-builder/src/components/Dashboard.tsx
@@ -92,7 +92,7 @@ const Dashboard = () => {
     }
 
     try {
-      await sdk.txs.send({ txs: transactions.map((d) => d.raw) }).catch(console.error);
+      await sdk.txs.send({ txs: transactions.map((transaction) => transaction.raw) }).catch(console.error);
       setTransactions([]);
     } catch (e) {
       console.error('Error sending transactions:', e);
--- a/apps/tx-builder/src/components/Dashboard.tsx
+++ b/apps/tx-builder/src/components/Dashboard.tsx
@@ -91,8 +91,12 @@ const Dashboard = () => {
       return;
     }
 
-    await sdk.txs.send({ txs: transactions.map((d) => d.raw) }).catch(console.error);
-    setTransactions([]);
+    try {
+      await sdk.txs.send({ txs: transactions.map((d) => d.raw) }).catch(console.error);
+      setTransactions([]);
+    } catch (e) {
+      console.error('Error sending transactions:', e);
+    }
   }, [sdk.txs, transactions]);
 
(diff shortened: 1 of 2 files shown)


apps/tx-builder/src/components/Dashboard.tsx
```diff
@@ -69,6 +70,35 @@ const Dashboard = () => {
...
+    [transactions],
+  );
+
+  const handleSubmitTransactions = useCallback(async () => {
+    if (!transactions.length) {
+      return;
+    }
+
+    try {
+      await sdk.txs.send({ txs: transactions.map((d) => d.raw) }).catch(console.error);
```
dasanra, line 95: As convention we use full names even for this cases. I supose this is a transaction but as I didn't check in deep is not clear to me with only `d`

Don't worry to use longer names

dasanra (changes requested):

apps/tx-builder/src/components/Dashboard.tsx
yagopv: I was moving previous code around and didn't see this. Done!!

dasanra (approved):
A really useful improvement! Thank you!

JagoFigueroa:
All good, thank you Yago ;)

PR #181, opened 2021-11-05
[TX Builder] encodeFunctionCall not working with boolean values properly

### LINKED ISSUE #155: [TX-Builder] Bool parameters being forced to true on contract interactions
# Description

Noticed a bug with the transaction builder, where when interacting with a contract that has a bool parameter it sets it to True even though setting it as False in the input field. (Also tried setting the bool as 0 and swapped it to True as well)

## Environment

+ Browser: Chrome
+ Wallet: MetaMask
+ Safe: 0x19B3Eb3Af5D93b77a5619b047De0EED7115A19e7
+ Environment:
    + production (mainnet)

### Adding tx through transaction builder

Screen Shot 2021-08-12 at 8 18 47 PM

### Preview before confirming tx

Screen Shot 2021-08-12 at 8 18 55 PM

### Last Preview screen right before submitting tx
Screen Shot 2021-08-12 at 8 19 04 PM

yagopv:
## What it solves
Resolves #155 

## How this PR fixes it
Calling the `encodeFunctionCall` method with a string (containing a bool) value cause the encoding to always convert to zeroes so the final transaction always show "True". Seems to be a problem with the typing in `web3` as the signature for the method is:

```
encodeFunctionCall(abiItem: AbiItem, params: string[]): string;
```

But if you send a bool value it works just fine. We are forcing the conversion and is fixed

## How to test it
1) Open Tx Builder inside a safe
2) Follow the steps in #155 and add several transactions with true and false values
3) Submit and check in the final bundle that the values were converted right

Test contract for mainnet: `0xef0881ec094552b2e128cf945ef17a6752b4ec5d`

### DIFF
--- a/apps/tx-builder/src/components/Builder.tsx
+++ b/apps/tx-builder/src/components/Builder.tsx
@@ -57,7 +57,7 @@ const getInputHelper = (input: any) => {
 const paramTypeNumber = new RegExp(/^(u?int)([0-9]*)$/);
 
 // This function is used to apply some parsing to some value types
-const parseInputValue = (input: any, value: string): string | boolean => {
+const parseInputValue = (input: any, value: string): any => {
   // If there is a match with this regular expression we get an array value like the following
   // ex: ['uint16', 'uint', '16']. If no match, null is returned
   const isNumberInput = paramTypeNumber.test(input.type);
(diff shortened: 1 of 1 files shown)


dasanra (approved):

JagoFigueroa:
Hey there! I think I found a bit of an edge case when more than one boolean field is involved in one trx:
Screenshot 2021-11-08 at 12 28 50:

Check that the first boolean field is set to 'False' and the second to 'True'.

Well, confirmation screen is showing both fields as they were set to 'False':
Screenshot 2021-11-08 at 12 29 09

JagoFigueroa:
All good now. Gracias Yago!

PR #191, opened 2021-11-15
Disconnect WalletConnect session when changing safes

### LINKED ISSUE #160: [WalletConnect] If you switch safes, walletconnect keeps the same session
## Description
When connecting a Dapp through wallet connect Safe App it creates a session between the dapp and the current safe. When changing the safe this is not detected an we still use the same wallet connect session that is stored in the localStorage, even it should be tied to the previous safe address.

## Steps to reproduce:
1. Connect via Safe 1 to any dApp that supports walletconnect
2. Switch to Safe 2 and navigate to wc safe app
3. Safe App displays the session from Safe 1

## Expected results
Sessions are determined by wallet address

### Obtained result
Session is used by both safes

yagopv:
## What it solves
Resolves #160 

## How this PR fixes it
We now are storing independent sessions for each safe address. When the application starts we look for a session in the localStorage to bring the connection live.

We are using a different way to start the `WalletConnect` instance using `session` or `uri` and sending an optional `storageId` prop for allowing the walletconnect client to store the sessions separate.

## How to test it
1) Open different DApps in different browsers (or mobile apps). The DApps should allow to use WalletConnect. For example you can use:
   - [Uniswap](https://app.uniswap.org/#/swap)
   - [SushiSwap](https://app.sushi.com/es/swap)
   - [Cowswap](https://cowswap.exchange/#/swap)
2) Open different instances of the safe WalletConnect app using different safe addreses
3) Copy the QRs codes and paste them in the safe WalletConnect App. It should create a session with the chosen DApp
4) Change safes and see that the connection is being mantained
5) Disconnect, change safes, reconnect ... and check the sessions are being managed properly

### DIFF
--- a/apps/wallet-connect/src/hooks/useWalletConnect.tsx
+++ b/apps/wallet-connect/src/hooks/useWalletConnect.tsx
@@ -1,6 +1,7 @@
 import { useState, useCallback, useEffect } from 'react';
 import WalletConnect from '@walletconnect/client';
 import { IClientMeta } from '@walletconnect/types';
+import Connector from '@walletconnect/core';
 import { useSafeAppsSDK } from '@gnosis.pm/safe-apps-react-sdk';
 import { isMetaTxArray } from '../utils/transactions';
 import { areStringsEqual } from '../utils/strings';
@@ -142,6 +143,15 @@ const useWalletConnect = () => {
       if (uri) wcConnect(uri);
     }
   }, [connector, wcConnect]);
+
+  useEffect(() => {
+    if (connector) {
+      const { accounts } = (connector as Connector) || [];
+      if (accounts.length && !accounts.includes(safe.safeAddress)) {
+        wcDisconnect();
+      }
+    }
+  }, [connector, safe.safeAddress, wcDisconnect]);
 
   return { wcClientData, wcConnect, wcDisconnect };
 };
--- a/apps/wallet-connect/src/hooks/useWalletConnect.tsx
+++ b/apps/wallet-connect/src/hooks/useWalletConnect.tsx
@@ -1,6 +1,6 @@
 import { useState, useCallback, useEffect, useRef } from 'react';
 import WalletConnect from '@walletconnect/client';
-import { IClientMeta } from '@walletconnect/types';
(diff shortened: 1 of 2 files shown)


apps/wallet-connect/src/hooks/useWalletConnect.tsx
```diff
@@ -143,6 +144,15 @@ const useWalletConnect = () => {
...
 
+  useEffect(() => {
+    if (connector) {
+      const { accounts } = (connector as Connector) || [];
+      if (accounts.length && !accounts.includes(safe.safeAddress)) {
+        wcDisconnect();
+      }
+    }
+  }, [connector, safe.safeAddress, wcDisconnect]);
+
```
mikheevm, line 155: WalletConnect has an undocumented `storageId` option which is essentially the key to store the session at:
https://github.com/WalletConnect/walletconnect-monorepo/blob/v1.0/packages/helpers/types/index.d.ts#L243
https://github.com/WalletConnect/walletconnect-monorepo/blob/v1.0/packages/clients/core/src/storage.ts#L7

How about using this instead? I think it would be more convenient for those who manage multiple Safes.

yagopv: mmm, and how do you want to use this? This is the id for the current session isn't it? Then when changing safes should be the same i guess. Could you elaborate on this a little more 🙃

mikheevm: You can include the wallet address to the key the session is stored

yagopv: so you mean storing multiple sessions and try to get one if the safe address match?

mikheevm: exactly

yagopv: Changed the approach and now seems to manage several safes nicely. Thanks 😁

mikheevm (approved):

JagoFigueroa:
Working like a charm, thank you guys ;)

PR #193, opened 2021-11-16
[TX Builder] Added Address Input Component to TX Builder

### LINKED ISSUE #163: [TX Builder] Enhance the address input
* [ ] Use eth.getCode to validate that the address is a contract (it will return 0x if not)
* [ ] Support EIP-3770 (prefixed addresses)

DaniSomoza:
## What it solves
Resolves #163

Address Input Component to TX-Builder Safe App

## Screenshots

Captura de pantalla 2021-11-16 a las 16 36 09
Captura de pantalla 2021-11-16 a las 16 36 20

### DIFF
--- a/apps/tx-builder/src/components/Dashboard.tsx
+++ b/apps/tx-builder/src/components/Dashboard.tsx
@@ -1,7 +1,9 @@
-import { Text, Title, Link, TextField } from '@gnosis.pm/safe-react-components';
-import React, { useState, useCallback } from 'react';
+import { Text, Title, Link, AddressInput } from '@gnosis.pm/safe-react-components';
+import React, { useState, useCallback, useEffect } from 'react';
 import { useSafeAppsSDK } from '@gnosis.pm/safe-apps-react-sdk';
 import styled from 'styled-components';
+import { InputAdornment } from '@material-ui/core';
+import CheckCircle from '@material-ui/icons/CheckCircle';
 
 import { ContractInterface } from '../hooks/useServices/interfaceRepository';
 import useServices from '../hooks/useServices';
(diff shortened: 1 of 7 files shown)


DaniSomoza:

apps/tx-builder/src/components/Dashboard.tsx
```diff
@@ -99,6 +74,47 @@ const Dashboard = () => {
...
+    (address: string | null) => {
+      if (!address) {
+        return false;
+      }
+      return services?.web3?.utils.isAddress(address);
+    },
+    [services.web3],
+  );
+
+  const hasError = !isValidAddress(addressOrAbi) && !contract;
```
dasanra, line 116: I would call it `showBuilder` and remove the three !

apps/tx-builder/src/components/Dashboard.tsx
```diff
@@ -99,6 +74,47 @@ const Dashboard = () => {
...
+        setLoadContractError('No ABI found for this address');
+        console.error(e);
+      }
+      setIsABILoading(false);
+    }
+
+    loadContract(addressOrAbi);
+  }, [addressOrAbi, services.interfaceRepo, services.web3]);
+
+  function getAddressFromDomain(name: string): Promise<string> {
```
dasanra, line 102: Mixed use of arrow and normal functions in this file too

package.json
```diff
@@ -18,6 +18,8 @@
   "dependencies": {
     "@gnosis.pm/safe-apps-react-sdk": "4.0.5",
     "@gnosis.pm/safe-apps-sdk": "5.0.0",
+    "@gnosis.pm/safe-react-components": "git://github.com/gnosis/safe-react-components.git#27ecf08b4a9fbd6d81be6d550bbdfc92eb632d68",
```
dasanra, line 21: We should be careful with this move. Currently all safe apps are using different versions of `safe-react-components`. If we move this here an update could lead to a full check of all the other apps

mikheevm: In general, I think we should restrict this practice of using git commit hashes to development only. If you remove this hash from GitHub (let's say remove a branch), you won't install dependencies anymore.

apps/tx-builder/src/components/Dashboard.tsx
DaniSomoza: done! I called it `isValidAddressOrContract`, what is your opinion about it?

apps/tx-builder/src/components/Dashboard.tsx
DaniSomoza: done!

dasanra (approved):

JagoFigueroa:
Hi there! I've noticed that when adding a smart contract address from a different network no validation is performed in terms of notifying the user that it is an invalid address:

Old implementation:
old

New implementation:
new

I think we should keep that validation @DaniSomoza  :)

Also, I find a bit weird that when copy / pasting  an address without a chain prefix it is automatically added in the 'to address' field. This is something very minor though.

question

JagoFigueroa:
All good now, thank you @DaniSomoza ;)

JagoFigueroa:
Moving this task back to QA as I will give it a second go after new changes are approved :)

JagoFigueroa:
All bueno, thank you @DaniSomoza 

I think this might be you right now
unnamed

PR #196, opened 2021-11-17
Make the transaction detail in the confirm modal break to next line for avoiding overflow

### LINKED ISSUE #194: [Tx Builder] Methods with many parameters overflow the review modal
## Description
When using a method with many long parameters, for example many addresses, we don't break line. This can make the modal to overflow the window not showing the user the confirm button
image

## Steps to reproduce
 1. We got an example smart contract on xDai
 2. Use it in the Transaction builder https://blockscout.com/xdai/mainnet/address/0x43d13D7B83607F14335cF2cB75E87dA369D056c7/write-contract
 3. Fill the data in the batchMint function. There is an example image below
 4. 
## Expected result
image

yagopv:
## What it solves
Resolves #194 

## How this PR fixes it
Adding a CSS rule for avoiding overflow. We as well added a little margin to separate transactions

## How to test it
1) Open Transaction builder and add some transactions using the contract indicated in the issue
2) Click the send transactions button
3) In the modal, transactions should be visible and not overflowed and the buttons should be clickable

### DIFF
--- a/apps/tx-builder/src/components/ModalBody.tsx
+++ b/apps/tx-builder/src/components/ModalBody.tsx
@@ -1,9 +1,15 @@
-import { Button, Text } from '@gnosis.pm/safe-react-components';
 import React from 'react';
+import { Button, Text } from '@gnosis.pm/safe-react-components';
 import Box from '@material-ui/core/Box';
+import styled from 'styled-components';
 import { ProposedTransaction } from '../typings/models';
 
 type Props = { txs: Array<ProposedTransaction>; deleteTx: (index: number) => void };
+
+const WrappedText = styled(Text)`
+  word-break: break-all;
+  margin: 5px 0;
+`;
 
 export const ModalBody = ({ txs, deleteTx }: Props) => {
   return (
@@ -20,7 +26,7 @@ export const ModalBody = ({ txs, deleteTx }: Props) => {
           <Button size="md" variant="outlined" iconType="delete" color="error" onClick={() => deleteTx(index)}>
             {''}
           </Button>
-          <Text size="lg">{tx.description}</Text>
+          <WrappedText size="lg">{tx.description}</WrappedText>
         </Box>
       ))}
     </>


apps/tx-builder/src/components/ModalBody.tsx
```diff
@@ -1,10 +1,16 @@
...
+import { Button, Text } from '@gnosis.pm/safe-react-components';
 import Box from '@material-ui/core/Box';
+import styled from 'styled-components';
 import { ProposedTransaction } from '../typings/models';
 
 type Props = { txs: Array<ProposedTransaction>; deleteTx: (index: number) => void };
 
+const WrappedText = styled(Text)`
+  word-break: break-all;
+  margin: 5px 0;
```
mikheevm, line 11: should we use spacings from the theme instead of hardcoded values?

yagopv: Good idea, i can do it. BTW this margin/paddings are mostly hardcoded in all the apps  so perhaps we should add an improvement for cleaning this type of things

yagopv: So, I think is better to leave it like this as in all the safe apps there are fixed values instead of using the ones in the theme. We are using as well a direct commit in this app instead the version and we don't have access to the `margin` prop for using predefined spacing

https://github.com/gnosis/safe-react-apps/blob/development/apps/tx-builder/package.json#L7

So, need to update but ... the latest version have a bug with the checkboxes as well being addressed currently but not yet released.

My suggestion would be to resolve this PR with the fixed value and create a ticket for updating the `safe-react-components` package to the latest version and use the theme properly instead fixed values. What do you think @dasanra ?

dasanra: For me it's correct right now to resolve this issue like this and to improve the `safe-react-components` to provide theming.

Bigger break in `safe-react-components` can be caused on web. External apps are not using this library in a very heavy way, although we should make clear this changes in the changelog and with semantic versioning

dasanra (approved):

JagoFigueroa:
Looking much better now and I can see that it is responsive as well, thank you guys ;)

Screenshot 2021-11-19 at 09 03 53

PR #203, opened 2021-11-19
[Drain Safe] Handle exclusions improvements

### LINKED ISSUE #201: [Drain Safe] Change exclude logic
## Overview
I had some insights from design, and our exclude pattern raised some concerns. Is not really logic for the user to check an item to be excluded so the suggestion is that we do the inverse logic.

## Goals
Improve include/exclude UX in Drain Safe

## Requirements
1. Change "Exclude" column to "Transfer" where all tokens are selected by default
2. Use will uncheck those tokens which doesn't want to be transferred
3. When not all assets are check transfer button should change from "Transfer everything" to "Transfer X assets" being X the number of assets
4. When the user unchecks all assets, the transfer button should be disabled
5. ~~Ideally check and uncheck could be done by clicking the whole row (if the current table allows this behavior)~~

yagopv:
## What it solves
Resolves #201 

## How this PR fixes it
We are changing the UX for excluding tokens. Instead selecting what to exclude now we are going to see a transfer column with all the assets selected by default. Then we can exclude assets by unselecting the corresponding checkbox in the asset row.

The transfer button is enhance as well showing the current status of the selection (Everything or N assets) and is able to detect no assets selected and change to a disable status.

## How to test it
Open a safe-web with the corresponding custom app pointing to the pr deployed url

### DIFF
--- a/apps/drain-safe/src/GlobalStyle.ts
+++ b/apps/drain-safe/src/GlobalStyle.ts
@@ -23,6 +23,10 @@ const GlobalStyle = createGlobalStyle`
         width: 100% !important;
     }
 
+    tbody > tr {
+        cursor: pointer;
+    }
+
     @font-face {
         font-family: 'Averta';
         src: local('Averta'), local('Averta Bold'),
--- a/apps/drain-safe/src/GlobalStyle.ts
+++ b/apps/drain-safe/src/GlobalStyle.ts
@@ -23,10 +23,6 @@ const GlobalStyle = createGlobalStyle`
         width: 100% !important;
     }
 
-    tbody > tr {
-        cursor: pointer;
-    }
-
     @font-face {
         font-family: 'Averta';
         src: local('Averta'), local('Averta Bold'),
--- a/apps/drain-safe/src/__tests__/App.test.js
+++ b/apps/drain-safe/src/__tests__/App.test.js
@@ -98,7 +98,7 @@ describe('<App />', () => {
   });
 
   it('should allow to order token by numeric prop', async () => {
-    const { container, debug } = renderWithProviders(<App />);
+    const { container } = renderWithProviders(<App />);
 
     await screen.findByText(/chainLink token/i);
     const amountColumnHeaderElement = screen.getByText(/amount/i);
--- a/apps/drain-safe/src/__tests__/App.test.js
+++ b/apps/drain-safe/src/__tests__/App.test.js
@@ -42,7 +42,7 @@ describe('<App />', () => {
     const { container } = renderWithProviders(<App />);
     const { sdk } = useSafeAppsSDK();
 
-    await screen.findByText(/chainLink token/i);
+    await screen.findByText(/chainlink token/i);
     fireEvent.change(screen.getByRole('textbox'), { target: { value: '0x301812eb4c89766875eFe61460f7a8bBC0CadB96' } });
     fireEvent.click(screen.getByText(/transfer everything/i));
     await waitFor(() => expect(sdk.txs.send).toHaveBeenCalledWith(mockTxsRequest));
@@ -57,7 +57,7 @@ describe('<App />', () => {
     fireEvent.click(checkboxes[3]);
     fireEvent.click(checkboxes[4]);
     fireEvent.change(screen.getByRole('textbox'), { target: { value: '0x301812eb4c89766875eFe61460f7a8bBC0CadB96' } });
-    fireEvent.click(screen.getByText(/transfer everything/i));
+    fireEvent.click(screen.getByText(/transfer 2 assets/i));
 
     await waitFor(() =>
       expect(sdk.txs.send).toHaveBeenCalledWith({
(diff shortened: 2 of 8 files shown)


apps/drain-safe/src/__tests__/App.test.js
```diff
@@ -78,32 +78,44 @@ describe('<App />', () => {
...
+    await waitFor(() => {
+      const tableRows = document.querySelectorAll('tbody tr');
+      expect(within(tableRows[4]).getByText(/chainlink token/i)).toBeDefined();
+      expect(within(tableRows[0]).getByText(/uniswap/i)).toBeDefined();
+    });
   });
 
   it('should allow to order token by numeric prop', async () => {
-    const { container } = renderWithProviders(<App />);
+    const { container, debug } = renderWithProviders(<App />);
```
mikheevm, line 101:
```suggestion
    const { container } = renderWithProviders(<App />);
```

apps/drain-safe/src/components/Balances.tsx
```diff
@@ -104,6 +105,7 @@ function Balances({
       sortedByHeaderId={orderBy}
       sortDirection={order}
       onHeaderClick={handleHeaderClick}
+      onRowClick={handleExclusion}
```
mikheevm, line 108: TableRow is not an interactive element, and we shouldn't add any event listeners to it. This is inaccessible.

https://stackoverflow.com/questions/64776155/make-table-row-or-cell-be-read-by-a-table-as-clickable

mikheevm (changes requested):

apps/drain-safe/src/components/Balances.tsx
yagopv: I'm just using the safe component library @mikheevm and is the requirement in this US

yagopv: Perhaps is this is really important should be addressed in the library but a row click event i think is no that strange. Suggestions in stackoverflow i think overcomplicate this

mikheevm: This is inaccessible. People that use screen readers to navigate through the web page won't be able to use this.
I suggest you read this: https://developer.mozilla.org/en-US/docs/Learn/Accessibility/What_is_accessibility.

> Suggestions in stackoverflow i think overcomplicate this

What's the overcomplication in making a button?

mikheevm: I created a [ticket](https://github.com/gnosis/safe-react-components/issues/169) in the component library

yagopv: I don't mean adding a button is complicated. I'm not considering it because the requirements in this ticket are to add the click in the row level. I'm referring to the other solutions in the link to stackoverflow you provided. 

Anyway, we can reformulate the requirement of course @dasanra

dasanra: @yagopv We can consider this for later. Was an appreciation from designers, but as I wrote in the ticket not a hard requirement if there is a good tech reason.

I think this was thought as for mobile devices, where you can click a full row to select or unselect more easily than having to pick a small checkbox

dasanra: @yagopv @mikheevm I found after some investigation that Material UI already provides a component to solve this issue
https://mui.com/components/tables/#data-table

This table is not using table html components and is much more convenient for modern table uses. We can consider if it may be worth to migrate to this kind of table instead traditional structure, that is not that friendly to accessibility.

But anyway to close this ticket I would just remove this requirement and add all the other improvements

mikheevm: I checked the implementation out of curiosity: https://github.com/mui-org/material-ui-x/blob/e45cf31723303ccc20cf21cccb6ef1e975364bed/packages/grid/_modules_/grid/components/GridRow.tsx#L225 🤯 

Making such a component accessible is a lot of work; developing a component library is no joke

yagopv: ok removed the row event!!

yagopv: If we want truly accesible tables i think we need a complete solution like the mui one @dasanra commented because if you want to navigate this table using a screen reader I think is is completely imposible as roles and indexes are missing

mikheevm: We don't have to "want" accessibility. It should be included by default.

apps/drain-safe/src/GlobalStyle.ts
```diff
@@ -23,6 +23,10 @@ const GlobalStyle = createGlobalStyle`
         width: 100% !important;
     }
 
+    tbody > tr {
+        cursor: pointer;
+    }
```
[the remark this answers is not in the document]
yagopv: nope!!

mikheevm (approved):
🎉

JagoFigueroa:
All good from my end, thank you ;)

Updating related test cases as well.

PR #211, opened 2021-11-24
[TX BUILDER] Use transaction service for centralizing contract info retrieval

### LINKED ISSUE #169: Transaction Builder, check ABI availability
## Description

Something is strange on the transaction builder, well-used addresses like DAI and USDC will return when their addresses are entered into the tx builder, an "ABI not found" message. 

The strange thing is that this is not consistent, entering in the same contract address a few minutes later will return the function methods for the contract. 

## Steps to reproduce
For reference, please use the DAI contract: 
https://etherscan.io/address/0x6b175474e89094c44da98b954eedeac495271d0f

## Expected result
Smart contract ABI is correctly fetch from Etherscan

## Obtained result
Currently in some moments the ABI is not returned. We should check if it's related to network speed issues or missing Etherscan API key in TX Builder configuration

yagopv:
## What it solves
Resolves #169 

## How this PR fixes it
We are adding a new way to get the abis as the current one is limited to 5 requests per seconds or adding API keys from the corresponding providers.

The new way about calling:
1) [Sourcify](https://sourcify.dev/) api
2) Transaction service ([new endpoint](https://github.com/gnosis/safe-client-gateway/issues/700) through gateway)

We are going to make this calls sequentially so if the contract is not found in Sourcify then we will use the transaction service as fallback

Currently we are using a direct call to the backend services for 2) as the backend team is implementing the changes

## How to test
1) Enter a contract using TX Builder
2) Check we call sourcify ...
```
// You can get contracts published in sourcify using this request
https://sourcify.dev/server/files/contracts/{chainId}
```
3) ... and if not found then the transaction service
```
Use Safe contract for example
0xd9Db270c1B5E3Bd161E8c8503c55cEABeE709552
```


DaniSomoza:
For me I tried to use Safe Addresses (`rin:0x57CB13cbef735FbDD65f5f2866638c546464E45F`) as a destination address and it seems that both services are returning a 404 error. I am not sure if that its expected or not:

Captura de pantalla 2021-12-02 a las 16 12 12
Captura de pantalla 2021-12-02 a las 16 12 22

For other Smart contracts (like `0x8DB05dFae0DDAC842692Ae6193Ce449D48961596` and `0x49d4450977E2c95362C13D3a31a09311E0Ea26A6`) Sourcify api is returning 404 and Transaction service is working as expected ( I think that this is expected because those smart contracts are not verified in Sourcify)

yagopv:
Yes checked that one. Is working fine but not found using both services

dasanra:
Here you have an example of the new route using the `client-gateway`

https://safe-client.staging.gnosisdev.com/v1/chains/4/contracts/0x00000000000045166C45aF0FC6E4Cf31D9E14B9A

JagoFigueroa:
I can confirm that it is working as expected :)

The problem I see with the new approach is that the message 'No ABI found for this address' will be more present for the app as resolution is going to decrease dramatically (specially on Rinkeby) but nothing we can really do about this at the moment.

JagoFigueroa (approved):

Issue #214, opened 2021-11-25
[TX-Builder] Allow custom contract data


dasanra:
## Overview
Currently in the Transaction Builder we only allow to create new transactions by selecting and filling the complete information after an ABI is provided and all fields are filled. Some users could prepare some custom data in some other dapp and just be willing to send it from the TX-Builder, for example sending a transaction where the signatures were gathered off chain

## Goals
We should allow switching to show custom hex encoded data the same way we allow it in contract interaction, either converting the user input to hex data or allowing to send custom hex data

## Requirements
- [ ] Allow input custom hex data to add a new transaction to the transaction builder
- [ ] Show custom hex data when user filled the different inputs when interacting with a smart contract

## Screens
Example in "Contract interaction" used as TX-Builder
![image](https://user-images.githubusercontent.com/25051234/143454190-7c1ae59a-429e-4816-952c-db1921d4bc61.png)

Example in "Contract interaction" of custom hex data transaction
![imagen](https://user-images.githubusercontent.com/25051234/143482260-0a2cedad-98eb-438a-a4dc-99a29f9d2b0b.png)

Any doubts we should check contract interaction feature on Safe web

JagoFigueroa:
Hi there! Just to clarify @dasanra custom data is always expected to come with an {{asset}} value field regardless of the method used right? 

@yagopv made it so that instead of a required field it is now optional but it is different to the current implementation on web (which I really like but I mention this in case needs to be exactly the same. When no value is input, it will be understood as a 0 instead of being a required 0.

Current implementation when performing a contract interaction on web:

![Screenshot 2021-12-13 at 11 03 25](https://user-images.githubusercontent.com/92332710/145792321-9aaa9339-6aaf-4975-b6f5-f0ebbd1ce959.png)

Yagos approach on tx builder: 

![Screenshot 2021-12-13 at 11 03 48](https://user-images.githubusercontent.com/92332710/145792406-a31d3267-cb24-40c5-afb1-b9f5df252896.png)

PR #220, opened 2021-11-29
[TX BUILDER] Add input validation before adding transactions

### LINKED ISSUE #195: [Tx Builder] Don't allow to perform transactions with negative values
## Description
We should not allow to add an address to address transaction when the {{asset}} value field is negative

## Steps to reproduce
Go to Transaction Builder app and add a new address to address transaction with negative value:
 

Screenshot 2021-12-01 at 09 49 46

## Expected result
Frontend validation blocks the user of adding a negative transaction

## Obtained result
Transaction(s) can be added and executed.

yagopv:
## What it solves
Resolves #195 

## How this PR fixes it
Adding validation for the input value before adding the transaction

## How to test it
1) Add a contract with no abi
2) Enter incorrect values like characters, negatives ... and check you cannot add a transaction after pressing the "Add transaction" button
3) Add correct values and check you can add transactions

### DIFF
--- a/apps/tx-builder/src/components/Builder.tsx
+++ b/apps/tx-builder/src/components/Builder.tsx
@@ -125,7 +125,7 @@ export const Builder = ({
 
   const isInputValueValid = useCallback((val: string) => {
     const value = Number(val);
-    if (isNaN(value) || value <= 0) {
+    if (isNaN(value) || value < 0) {
       return false;
     }
 
--- a/apps/tx-builder/src/components/Builder.tsx
+++ b/apps/tx-builder/src/components/Builder.tsx
@@ -123,6 +123,15 @@ export const Builder = ({
     setInputCache(inputCache.slice());
   };
 
+  const isInputValueValid = useCallback((val: string) => {
+    const value = Number(val);
+    if (isNaN(value) || value <= 0) {
+      return false;
+    }
+
+    return true;
+  }, []);
+
   const getContractMethod = useCallback(() => contract?.methods[selectedMethodIndex], [contract, selectedMethodIndex]);
 
   const handleSubmit = () => {
@@ -140,6 +149,11 @@ export const Builder = ({
     const web3 = services.web3;
 
     if (!web3) {
+      return;
+    }
+
+    if (!isInputValueValid(valueInput)) {
+      setValueError(`${nativeCurrencySymbol} value`);
       return;
     }
 
@@ -206,8 +220,7 @@ export const Builder = ({
 
   const onValueInputChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     setValueError(undefined);
-    const value = Number(e.target.value);
-    if (isNaN(value) || value < 0) {
+    if (!isInputValueValid(e.target.value)) {
       setValueError(`${nativeCurrencySymbol} value`);
     }
     setValueInput(e.target.value);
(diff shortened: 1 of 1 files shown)


mikheevm:
Hey, I'm not sure if you saw [my comment](https://github.com/gnosis/safe-react-apps/issues/195#issuecomment-981614117), but in short, there are certain cases where transactions with 0 value may be helpful, and it's not clear whether this is a bug. So IMO we should ask for product/team opinion.

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -143,6 +152,11 @@ export const Builder = ({
       return;
     }
 
+    if (!isInputValueValid(valueInput)) {
+      setValueError(`${nativeCurrencySymbol} value`);
+      return;
+    }
+
```
dasanra, line 159: Wouldn't it make more sense to disable the button?

Shouldn't it be already showing the error value, that should be set in the function `onValueInputChange`?

yagopv: I think is better to let the user make errors by pressing buttons. Disabling buttons i think is a little frustrating from a user perspective because sometimes you don't know what to do.

Besides that, If we would like to go that way perhaps we should do the same with the other form fields right? ... for keeping consistency

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -123,6 +123,15 @@ export const Builder = ({
     setInputCache(inputCache.slice());
   };
 
+  const isInputValueValid = useCallback((val: string) => {
```
mikheevm, line 126: You can define this function outside the component body and you won't need `useCallback`

yagopv: yes I can do it.

In this concrete component there are a lot of functions created inside and outside but there is no concrete pattern for doing it. What I'm doing in this issue #214  is extracting the functions from outside the component to the utils file. What do you think if I do the same with this?

mikheevm: I think it's OK to keep functions in a single file if you don't use them in other files. No strong preference, though.

> there is no concrete pattern for doing it

well, besides the pattern, using it with useCallback is less efficient because the memoization isn't free

yagopv: yes, with "pattern" i mean in this concrete component there are more useCallback uses that doesn't make sense for the same memoization issues you are mentioning and should be extracted as well. I will do it in the other issue i'm developing.

I extracted this function to outside anyway. Thanks !!

mikheevm (approved):
gracias

apps/tx-builder/src/components/Builder.tsx
dasanra: Not a strong opinion there, we can get over this again if necessary when we have design updates

dasanra (approved):

JagoFigueroa:
Input validation is working as expected. I was talking with @yagopv and we noticed that after the component changes for the address input implemented by @DaniSomoza  and his, now we have 3 different types of styling for the fields present in this app when front validation kicks in:

Screenshot 2021-12-02 at 15 09 07

Screenshot 2021-12-02 at 15 32 03

What do you think if I create a separate improvement to have this behaviour aligned for all fields @dasanra ?

dasanra:
@JagoFigueroa Perfect for me. We can close this ticket and work in a new one to align input validation styles

yagopv:
background colors are different as well. Going to merge as this is in develop already

Issue #235, opened 2021-12-09
Fetch RPC provider from the SDK


dasanra:
## Overview
As followup of https://github.com/gnosis/safe-apps-sdk/issues/254 we need to cleanup the manual configuration of RPC endpoints for each chain. We should get this information using the Apps SDK detecting the WebApp current network

## Goals
 - Remove chain dependency from TX-Builder and Drain Safe app

## Requirements
- [x] Create a PR in `safe-react` updating the communicator to return the complete chain configuration
- [x] Upgrade types in `safe-apps-sdk` in case is necessary
- [ ] Remove RPC endpoints from Tx-Builder and Drain Safe apps and get them from the SDK

DaniSomoza:
Added https://github.com/gnosis/safe-apps-sdk/pull/264 PR

DaniSomoza:
Added PR in `safe-react`:  https://github.com/gnosis/safe-react/pull/3164

PR #236, opened 2021-12-09
Added a warning icon if transfer cost is higher than token market value

### LINKED ISSUE #164: [Drain Safe] Warn for those tokens with really low amount or USD value
## Overview
Currently the user can select which tokens to be transferred when draining a Safe. But many safes suffer of spam tokens or may have tokens with low amounts which are not interesting to transfer, because the transfer cost would be higher than the token value.

## Goals
We should give the user some feedback to detect this tokens that may not be worth to transfer because the cost would be higher than the cost of moving them.

* [ ] We should compare for each token when the fiat value can overpass the transaction cost of transferring it
* [ ] Show a yellow warning sign :warning: or something similar next to the fiat value of the token.

## Requirements
How to check that transfer value is higher than the fiat value?
 - We can have an approximation of how much calling an ERC20 send will cost in terms of gas. This will be a fixed amount.
 - We should fetch the gwei fast price for the chain at that moment so we can say if it could be more costly than current value
 - We should fetch the fiat value of the token used to pay the transaction so we can compare the amount to the cost

DaniSomoza:
## What it solves
Resolves #164

## How this PR fixes it
Added a Warning Icon & Tooltip if transfer cost is higher than token market value.
Also Added a new `CurrencyCell` Component

## Screenshots
Captura de pantalla 2021-12-09 a las 14 23 57


apps/drain-safe/src/components/App.tsx
```diff
@@ -108,13 +114,52 @@ const App: React.FC = () => {
     }
   }, [balancesError]);
 
+  useEffect(() => {
```
yagopv, line 117: What do you think about creating a useWeb3 hook? for avoid passing down props to de final CurrencyCell. You can gather the chains.ts info there as well as we are going to remove it later

DaniSomoza: I think that it's a good idea to create a `useWeb3` hook, but I think that we should keep it in the `App.tsx` because we only want instantiate web3 and call `getGasPrice` only once and not in each `CurrencyCell`.

apps/drain-safe/src/utils/chains.ts
```diff
@@ -0,0 +1,31 @@
...
+  ENERGY_WEB_CHAIN = '246',
+  ARBITRUM = '42161',
+  VOLTA = '73799',
+}
+
+export const rpcUrlGetterByNetwork: {
+  [key in CHAINS]: null | ((token?: string) => string);
+} = {
+  [CHAINS.MAINNET]: (token) => `https://mainnet.infura.io/v3/${token}`,
+  [CHAINS.MORDEN]: null,
```
mikheevm, line 20: Do we want to include chains to this enum if we don't support them?

apps/drain-safe/src/components/App.tsx
mikheevm: +1 for a separate hook, I find that it vastly improves readability

yagopv: Could perhaps be a separate ticket @dasanra ? as TX Builder and Drain Safe can share the same hook and as @DaniSomoza mentioned me offline, we will need a context provider as well for avoid instance recreation.

mikheevm: Why would we want to avoid it? In the drain-safe app, it's only used in one component and then passed down via props. To me, it looks like just moving it to a separate file

yagopv: For avoid the prop to only exist to be sent down. In this case in the Balances I think

yagopv: but the context is fine isn't it ? I mean, a context for the web3 instance and a hook to instantiate it using our getChainInfo populated with RPC url (task in progress) could be helpful for retrieving it in any level of the component tree

apps/drain-safe/src/utils/chains.ts
dasanra: We are doing a copy 1:1 of what we have on TX-Builder. This is to be deprecated with [#235](https://github.com/gnosis/safe-react-apps/issues/235) in both apps

apps/drain-safe/src/components/App.tsx
DaniSomoza: done! Added useWeb3.js hook

apps/drain-safe/src/components/App.tsx
```diff
@@ -108,13 +113,28 @@ const App: React.FC = () => {
...
     <FormContainer onSubmit={onSubmit} onReset={onCancel}>
       <Flex>
         <Logo />
         <Title size="md">Drain Account</Title>
       </Flex>
-      <Balances assets={assets} exclude={excludedTokens} onExcludeChange={handleExcludeChange} />
+      <Balances
+        ethFiatPrice={ethFiatPrice}
+        gasPrice={gasPrice}
+        web3={web3}
```
yagopv, line 133: With the useWeb3 new hook you don't really need to pass it down right?

DaniSomoza: Yep, I decided to keep the `const { web3 } = useWeb3();` and `setGasPrice` useState + useEffect in the App.tsx, But maybe makes sense to declare it directly in the Balances Component

yagopv: Really in CurrencyCell. In Balances you don't need i think

DaniSomoza: but as we talked in the other thread we will need a context provider for avoid instance recreation. And we will address it in a different ticket.

yagopv: ahh right 😂 ... so let's improve it later

yagopv (approved):

PR #239, opened 2021-12-09
Use Select for bool fields

### LINKED ISSUE #184: Transaction Builder - Make boolean fields actually boolean
## Description
We should avoid using string type fields for boolean fields as it can lead to unnecessary edge cases as values different from 'True' or 'False' are accepted:

Screenshot 2021-11-08 at 12 48 41

For this, we should create/use the adequate front component so that it becomes a dropdown field with a true or false selector.

yagopv:
## What it solves
Resolves #184 

## How this PR fixes it
Using the `<Select />` component from the safe-react-components instead a regular input

## How to test it
1) Choose `True` and `False` values from the dropdown and add several transactions using them
2) Send transactions to safe-web and check that the values are correctly propagated

### DIFF
--- a/apps/tx-builder/src/components/Builder.tsx
+++ b/apps/tx-builder/src/components/Builder.tsx
@@ -110,6 +110,11 @@ const isInputValueValid = (val: string) => {
   return true;
 };
 
+const BOOLEAN_ITEMS = [
+  { id: 'true', label: 'True' },
+  { id: 'false', label: 'False' },
+];
+
 type Props = {
   contract: ContractInterface | null;
   to: string;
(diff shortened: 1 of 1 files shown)


DaniSomoza (approved):
Approved! :D

apps/tx-builder/src/components/Builder.tsx
```diff
@@ -281,6 +286,48 @@ export const Builder = ({
     }
   }, [transactions]);
 
+  const renderInput = (input: any, index: number) => {
```
mikheevm, line 289: I think we'd benefit more if this would be a separate component; IMO this `Builder.tsx` component is already long enough

yagopv: I agree with you but I decide to make the simplest change because I think is better to create separate components per field type (more or less like the AddressContractField one) with validation included ... and more independent because If you take a look to the AddressContractField there are props sent to it totally dependent on the Builder like the index position or the input itself, so no very decoupled al the end.

As we are going to have a ticket for implementing validation, I think it will be the right place to refactor this loop creating the individual decoupled components managing rendering and validation and being controlled by the builder.
 
So I felt that I was going to create a component that at the end we are going to change again shortly for something more elaborated

mikheevm (approved):

JagoFigueroa:
All good from my end, top job yago ;)

Issue #212, opened 2021-11-25
[Transaction Builder] Add network references to batched transactions view pop up


JagoFigueroa:
## Environment
 - Environment: any will do (also checked on EIP-3770 release branch https://pr3015--safereact.review-safe.gnosisdev.com/app/welcome)

- Tested after this PR https://github.com/gnosis/safe-react-apps/pull/193 adding new address input

## Steps to reproduce
 1. Go to transaction builder app
 2. Add at least 1 EOA to EOA transaction
 3. Click on Send transactions button

## Expected result
Add network reference.

## Obtained result
![Screenshot 2021-11-25 at 10 35 36](https://user-images.githubusercontent.com/92332710/143417944-89f2e24d-2af6-4ddf-9f57-05d43c37ef7a.png)

(for reference) on EIP-3770 release branch confirmation pop up references network on next step:

![Screenshot 2021-11-25 at 10 47 16](https://user-images.githubusercontent.com/92332710/143418376-bdd84dbc-8178-4641-b3bd-4ac138032a10.png)

PR #240, opened 2021-12-10
Feature/add new address input to drain safe

### LINKED ISSUE #198: [Drain Safe] Use new address input component
## Overview
Currently on Drain Safe we are using a custom address input. We should use the new address input component from `safe-react-components` that includes EIP-3770 support, validation and ENS name resolution.

Screenshot 2021-11-02 at 09 52 58

## Goals
Add new address input component with latest improvements

## Requirements
ENS should work on that chains that support it
 - Mainnet
 - Rinkeby

yagopv:
## What it solves
Resolves #198 

## How this PR fixes it
By updating the current input to the new `<AddressInput />` from the [@gnosis/safe-react-components](https://github.com/gnosis/safe-react-components) project

## How to test it
1) Enter address and make a transaction to validate all is working fine
2) Enter ENS and make a transaction to validate all is working fine

### DIFF
--- a/apps/drain-safe/package.json
+++ b/apps/drain-safe/package.json
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "private": true,
   "dependencies": {
-    "@gnosis.pm/safe-react-components": "^0.8.6",
+    "@gnosis.pm/safe-react-components": "^0.9.2",
     "@material-ui/core": "^4.12.3",
     "bignumber.js": "^9.0.1",
     "web3-eth-abi": "~1.6.0"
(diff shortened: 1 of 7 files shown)


mikheevm (approved):

JagoFigueroa:
Component is working as expected. I have noticed this:

https://user-images.githubusercontent.com/92332710/145776287-f84b9e45-8b1a-40ba-8dd9-d7995b94d72c.mov

@yagopv is checking if there is something we can do regarding the prefix being automatically added when a drain is cancelled or rejected for an address without one.

yagopv:
Right now we are showing the network prefix if available (always at the end) because we don't know the user config from the web. This is going to change when #212 will be finished

In the video you see it on reject because the component is rerendered and the prefix is available. I would wait for #212 to be resolved and continue with this like this as is only extra info for the user

JagoFigueroa:
Aaaaamigo. Then all good from my end ;)

PR #243, opened 2021-12-14
[TX Builder] Add method for check if we are ready for showing the input

### LINKED ISSUE #242: [Tx Builder] No 'To address' field when operating with ABIs
## Description
We need to add the 'To address' field when operating with ABIs as it is not possible to add any transaction without it:

Prod:

Screenshot 2021-12-14 at 13 23 00

Dev:

Screenshot 2021-12-14 at 13 24 06

yagopv:
## What it solves
Resolves #242 

## How this PR fixes it
We are adding a method for check if we should show de address field

## How to test it
1) Add addresses and check the destination address is prefilled
1) Add abis and check the destination address is showed empty


DaniSomoza (approved):
Approved!

PR #245, opened 2021-12-15
[Drain Safe] Update table component

yagopv:
## What it solves
Resolves https://github.com/gnosis/safe-react-components/issues/170

## How this PR fixes it
We are going to use a new component (https://v4.mui.com/components/data-grid)

By using this new component we can use a checkbox in the header for select/unselect all rows

The component is as well much more accesible when used with screen readers

## How to test it
1) Functionality of drain safe should be kept so a regression here is necessary
2) Check/Uncheck header for seeing all assets being selected/unselected
3) Use a screen reader as voice over to navigate through the cells and check/uncheck rows

### DIFF
--- a/apps/drain-safe/src/__tests__/App.test.js
+++ b/apps/drain-safe/src/__tests__/App.test.js
@@ -75,11 +75,13 @@ describe('<App />', () => {
   it('should drain the safe when submit button is clicked removing the tokens excluded by the user', async () => {
     renderWithProviders(<App />);
     const { sdk } = useSafeAppsSDK();
+
+    await screen.findByText(/chainlink token/i);
     const checkboxes = await screen.findAllByRole('checkbox');
 
-    fireEvent.click(checkboxes[1]);
-    fireEvent.click(checkboxes[3]);
+    fireEvent.click(checkboxes[2]);
     fireEvent.click(checkboxes[4]);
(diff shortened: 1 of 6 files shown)


apps/drain-safe/src/__tests__/App.test.js
```diff
@@ -146,11 +148,15 @@ describe('<App />', () => {
   it('Shows a Warning icon when token transfer cost is higher than its current market value ', async () => {
     renderWithProviders(<App />);
 
+    await screen.findByText(/maker/i);
+
     const warningTooltip =
       /Beware that the cost of this token transfer could be higher than its current market value \(Estimated transfer cost: /i;
 
+    const tableRows = document.querySelectorAll('.MuiDataGrid-row');
```
DaniSomoza, line 156: I think that this repeated line can be deleted

yagopv: yep!!, done

DaniSomoza (approved):
Approved! :D

JagoFigueroa:
So, weirdly enough I am not able to reproduce the issue with the empty row at the button regularly anymore (yago wasn't able as well so might have been an edge case) + the one at the upper side corner is expected due to the component itself.

Lets leave it to the experts if any design change were needed ;)

PR #246, opened 2021-12-17
Filter native tokens without value

### LINKED ISSUE #221: [Drain Account] Network token added to the list when amount is 0
## Overview
Network token is always added to the list even when its value is 0. If it isn't manually unchecked it will add a new transaction that will cost gas to the user

## Goals
If the network token (ETH, xDai, Matic...) value is 0 it shouldn't be added to the assets list. This way it also shouldn't be added to the drain transaction

## Requirements
- [ ] Avoid adding the network token to the list
- [ ] Check that it isn't added to the transaction to drain all funds

yagopv:
## What it solves
Resolves #221 

## How this PR fixes it
By filtering the native token when `fiatBalance` is zero
If there are no transferable tokens then we are showing a message to the user (See screenshots)

## How to test it
Test in several networks (mainnet, xDai ...)
1) Use a safe with a native token with no value and check that is no listed
2) Use a safe with a native token with value and check that is listed
3) Check the token is included/removed from the final transfer in the safe-web dialog
4) Check the message is showed instead the table if you don't have any transferable asset

### No assets message

image


DaniSomoza (approved):
Approved! :D

JagoFigueroa:
All good, thank you yago and a very nice touch to add a screen when there are no assets at all ;)

PR #248, opened 2021-12-20
Initial value using inputProps for destination address

### LINKED ISSUE #226: [Tx Builder] Persistent error when adding a valid transaction after an invalid one
## Description
An error is shown to the user when trying to add an invalid transaction to the batch but it shouldn't persist when a transaction is added successfully after it.

## Steps to reproduce
See attached video

## Expected result

Hide the error when user adds a transaction successfully to the batch of transactions

## Obtained result

https://user-images.githubusercontent.com/92332710/144435905-b6eb7649-832b-4a40-8c1b-c7b7d5de752f.mov

yagopv:
## What it solves
Resolves #226 

## How this PR fixes it
The concrete error specified in this issue was not happening anymore but other one arise. When removed the address value the field itself is removed so this PR fixed this other error instead:

https://user-images.githubusercontent.com/1576776/146756577-bfea4903-a753-4b95-932b-0c51afc15b75.mov

## How to test it
1) For the concrete error in this issue => Add an invalid transaction and see the error being showed (There was an error trying to add the transaction). Then add a valid address and click the Add transaction button to see that the error message is not showed anymore
2) Repeat the process but delete the value in the address field as showed in the video. The address field should be kept in the UI

### DIFF
--- a/apps/tx-builder/src/components/Builder.tsx
+++ b/apps/tx-builder/src/components/Builder.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, ReactElement, useCallback, useMemo, ChangeEvent } from 'react';
+import React, { useState, useEffect, ReactElement, useCallback, ChangeEvent } from 'react';
 import {
   Button,
   Text,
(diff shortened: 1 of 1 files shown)


DaniSomoza (approved):
Approved! :D

JagoFigueroa:
All good, muchas gracias caballeros ;)

PR #252, opened 2021-12-21
[Compound] Add COMP Claim support

### LINKED ISSUE #24: [Compound] Display COMP tokens earned and enable to claim
## Overview
As a user I want to see the amount of COMP tokens I have earned through interacting with the Compound protocol and be able to collect ("claim") these tokens.

Similar to how it works on: https://app.compound.finance

## Goals
 - The user should be able to claim COMP tokens generated
 - The user should be able to see the current APY in COMP tokens for any market

## Requirements
- [ ] Add a label showing currently owned COMP tokens and also a button to start the claim process. AFter clicking the button we create the transaction to claim all tokens
- [ ] Show current Distribution token APY the same way we show current interest rate

yagopv:
## What it solves
Resolves #24

## How this PR fixes it
1) By substituting current APR calculation for the Supply and Distribution APYs showed in the Compound site
2) Adding a section for claiming COMP

## How to test it
1) Test APY calculations against Compound site. Should work both Mainnet and Rinkeby for the Supply APY and only in mainnet for distribution tokens (Distribution APY and COMP collect)
2) Try to collect available COMP. Any cToken should be supplied in mainnet for start collecting. We are showing the 18 decimals of the COMP token and allowing collection but once tested we should allow only to claim COMP greater than 0.0001 as in the Compound site

## Screenshots

image


yagopv:
Perhaps I left something weird in the PR but as I'm going on holiday till next year i left this in code review for start gathering some feedback

apps/compound/src/hooks/useWeb3.ts
```diff
@@ -0,0 +1,75 @@
+import { useEffect, useState } from 'react';
+import Web3 from 'web3';
+import { useSafeAppsSDK } from '@gnosis.pm/safe-apps-react-sdk';
+
+function useWeb3() {
```
mikheevm, line 5: the return type is missing

apps/compound/src/components/WidgetWrapper.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from 'styled-components';
+
+type Props = {
+  children: React.ReactNode;
+};
+
+export default function WidgetWrapper({ children }: Props): React.ReactElement {
```
mikheevm, line 7: nitpick and a tip: if you use React.FC as a component type, it already includes type for `children`.

I'd say it's better to use arrow functions + React.FC just to stay consistent with how components are defined within the project.

apps/compound/src/components/InfoRow.tsx
```diff
@@ -0,0 +1,25 @@
+import { Divider, Text } from '@gnosis.pm/safe-react-components';
+import styled from 'styled-components';
+
+type Props = {
+  label: string;
+  data: string | React.ReactElement | undefined;
+};
+
+export default function InfoRow({ label, data }: Props): React.ReactElement | null {
```
mikheevm, line 9: it never returns null

apps/compound/src/abis/CompoundLens.ts
```diff
@@ -0,0 +1,373 @@
+const CompoundLensABI = [
```
mikheevm, line 1: I haven't looked into it but probably we can remove some methods from the abi we don't use to reduce the bundle size

apps/compound/src/components/CompBalance.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <InfoContainer>
+      <Title size="xs">Compound Token Balance</Title>
+      <InfoRow
+        label="Your COMP"
+        data={
+          <>
+            {Number.isNaN(balance) ? 0 : balance?.toFixed(18)}{' '}
+            <Link size="lg" onClick={handleOnCollect}>
+              Collect
+            </Link>
```
mikheevm, line 32: It's not semantic to use a link there, it should be a button (sorry if it's not your code)

apps/compound/src/App.tsx
```diff
@@ -1,57 +1,34 @@
...
-      return;
-    }
-    const chainId = safeInfo.chainId as CHAINS;
-    const web3Instance = new Web3(`https://${networkByChainId[chainId]}.infura.io/v3/${rpc_token}`);
-    setWeb3(web3Instance);
-  }, [safeInfo]);
+  const { web3 } = useWeb3();
+  const { sdk: appsSdk, safe: safeInfo, connected } = useSafeAppsSDK();
+  const { cTokenSupplyAPY, cDistributionTokenSupplyAPY, claimableComp, claimComp } = useComp(selectedToken);
+  const isMainnet = useRef(false);
```
mikheevm, line 31: I think `useMemo` suits better here

apps/compound/src/App.tsx
```diff
@@ -234,10 +199,15 @@ const CompoundWidget = () => {
...
         value: '0',
-        data: cTokenInstance.methods.redeemUnderlying(supplyParameter).encodeABI(),
+        data: cTokenInstance?.methods.redeemUnderlying(supplyParameter).encodeABI(),
       },
     ];
-    appsSdk.txs.send({ txs });
+
+    try {
+      await appsSdk.txs.send({ txs });
+    } catch {
```
mikheevm, line 208: would be helpful to log the actual error

apps/compound/src/App.tsx
```diff
@@ -201,30 +162,34 @@ const CompoundWidget = () => {
...
+          data: cTokenInstance?.methods.mint(supplyParameter).encodeABI(),
         },
       ];
     }
 
-    appsSdk.txs.send({ txs });
+    try {
+      await appsSdk.txs.send({ txs });
+    } catch {
+      console.error('Lock: Transaction rejected or failed');
```
mikheevm, line 186: would be helpful to log the actual error

apps/compound/src/components/WidgetWrapper.tsx
yagopv: nice!

apps/compound/src/App.tsx
yagopv: mmm why do you think so. I can change ir but not sure about the benefit. useMemo cause rerender

mikheevm: So, what you're doing now:
```
{isMainnet.current && (
            <InfoRow label="Distribution APY" data={cDistributionTokenSupplyAPY && `${cDistributionTokenSupplyAPY}%`} />
          )}
```

React doesn't track changes for `isMainnet.current` (that's why `useRef` is needed)
So, if isMainnet.current changes, the react won't update the output markup. And the reason why it works is because the app re-rendered after the value was set

yagopv:
Changed @mikheevm !!

apps/compound/src/abis/CompoundLens.ts
yagopv: I was adding full size abis. Yeah totally true. I left only the methods i need

apps/compound/src/App.tsx
yagopv: mmm right. I was thinking it as a load and forget variable because if user change network in safe web the ui is really reloading the web application and we need to enter the app again. But you are right let me change it for being future proof

yagopv: done, you can review it

mikheevm (approved):
impresionante trabajo señor

yagopv:
🤣 ... Love your spanish improvements

JagoFigueroa:
Looking good! Only testing missing from me would be to collect the COMPs but I don't have funding for mainnet. Waiting for finance team to do their magic ;)

mikheevm:
Can we merge this one and then test this functionality in https://github.com/gnosis/safe-react-apps/pull/261? Unfortunately, it's hard to review the other PR because it includes changes from this one.

JagoFigueroa:
If it's okay for @dasanra all good from my end :)

yagopv:
I think is a good idea

Issue #238, opened 2021-12-09
[TX Builder] Add minimum validation based on input type


dasanra:
## Overview
Currently we are not applying an exhaustive type check in the inputs, which can lead to some errors when we already know that some values are not compatible with the inputs. It also prevents us from showing a proper error to the user in the correct input field

## Goals
We should provide some basic validation at least when a field is required or it needs an explicit input type. Else we should show an error that give hints for the user if the value could be wrong

## Extra info
Here you can find a guide of types in Solidity
https://docs.soliditylang.org/en/v0.8.10/types.html#value-types

PR #256, opened 2022-01-03
[TX BUILDER] Align Field Styles

### LINKED ISSUE #228: [Tx Builder] Align styling for all available fields
## Description

Currently we are using at least 3 different implementations when it comes to validations / error handling for this app:

Screenshot 2021-12-02 at 15 09 07
Screenshot 2021-12-02 at 15 32 03

We should be able align this behaviour dynamically keeping in mind the huge amount of different fields and scenarios possible when it comes to smart contracts and ABIs.

yagopv:
## What it solves
Resolves #228

## How this PR fixes it
Most of the styles were fixed in other PRS
In this one we are adding "Invalid" in front of the "ETH value" label when the field has an error. This should be enough for aligning all the styles

## How to test it
1) Open TX Builder
2) Add a contract with no ABI for seeing the ETH value field. Enter a -1 and see how the label change to "Invalid ETH value"

### DIFF
--- a/apps/tx-builder/src/components/Builder.tsx
+++ b/apps/tx-builder/src/components/Builder.tsx
@@ -53,7 +53,7 @@ const StyledTextField = styled(TextFieldInput)`
 const StyledAddressInput = styled(AddressInput)`
   && {
     width: 520px;
-    margin-bottom: 10px;
+    margin-bottom: 1px;
 
     .MuiFormLabel-root {
       color: #0000008a;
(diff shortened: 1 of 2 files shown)


yagopv:
As a side note, we are going to add proper basic validation in the remaining fields here https://github.com/gnosis/safe-react-apps/issues/238

mikheevm (approved):
gmi

JagoFigueroa:
Styling is looking sharp, all good from my end. Opened this separate ticket affecting the same field https://github.com/gnosis/safe-react-apps/issues/263 as it is not directly related to this PR ;)

PR #262, opened 2022-01-10
Using web3-utils instead of web3 instance directly

### LINKED ISSUE #235: Fetch RPC provider from the SDK

DaniSomoza:
## What it solves
Resolves #235 

## How this PR fixes it
Now we are using `web3-utils` for utilities like `isAddress`, `isHexStrict`, `toChecksumAddress`, `toWei`, `fromWei`...
instead of using `web3` instance directly

we can not remove the hardcoded data `rpcUrl` by network in `useWeb3.ts` yet because we still need it for ENS resolution. This will be addressed in a different ticket.

## How to test it

Just check that Tx-Builder & Drain-safe works as expected


apps/tx-builder/src/components/Builder.tsx
```diff
@@ -266,15 +263,12 @@ export const Builder = ({
...
-  const isValidAddress = useCallback(
-    (address: string | null) => {
-      if (!address) {
-        return false;
-      }
-      return services?.web3?.utils.isAddress(address);
-    },
-    [services.web3],
-  );
+  const isValidAddress = useCallback((address: string | null) => {
```
mikheevm, line 266: this can be moved outside the component now because it doesn't depend on anything inside the component now

apps/tx-builder/src/components/Dashboard.tsx
```diff
@@ -81,21 +81,18 @@ const Dashboard = (): ReactElement => {
...
-  const isValidAddress = useCallback(
-    (address: string | null) => {
-      if (!address) {
-        return false;
-      }
-      return web3?.utils.isAddress(address);
-    },
-    [web3],
-  );
+  const isValidAddress = useCallback((address: string | null) => {
```
mikheevm, line 90: the same function exists in `Builder.tsx`

apps/tx-builder/src/components/Builder.tsx
DaniSomoza: nice catch! done! :D

apps/tx-builder/src/components/Dashboard.tsx
DaniSomoza: done [here](https://github.com/gnosis/safe-react-apps/pull/262#discussion_r785927409)! :D

yagopv (approved):

JagoFigueroa:
All good ;)

PR #266, opened 2022-01-11
[Compound] Get markets config from REST api 

### LINKED ISSUE #223: [Compound] Improve Compound markets detection
## Overview
Currently we are showing some outdated or deprecated markets in Compound safe app. Also we are not showing new markets with better yield. We should update the markets to allow users to interact with new markets and get better yields.

## Goals
- [ ] Automatically detect compound markets for example using [their graph](https://thegraph.com/hosted-service/subgraph/graphprotocol/compound-v2?query=All%20Markets)
- [ ] Avoid market duplication, deprecated markets should not be shown. Deprecated markets [can be check here](https://compound.finance/markets)

## Requirements
 - At the end of this ticket the market list should be up to date
 - Currently there is a fixed token list. This should be like this still for Rinkeby or we would break app compatibility with test chain

### LINKED ISSUE #270: [Compound] Hide non relevant tokens for Supply / Withdrawal
## Description

After the new implementation to fetch updated asset APYs https://github.com/gnosis/safe-react-apps/issues/223, we find that there are some tokens present in the API call but manually hidden in the Compound app (SAI & REP in mainnet). Maybe we should think of a way to hide these tokens as it is not possible to supply them on the app itself. + have 0 return.

Screenshot 2022-01-12 at 10 37 45

yagopv:
## What it solves
Resolves #223, resolves #270

## How this PR fixes it
We are retrieving the markets now from the Compound REST api. It only works on mainnet. For Rinkeby we fallback to some coins for testing purposes

## How to test it
1) Open Compound app (mainnet) and connect to your safe from https://app.compound.finance using WalletConnect
2) Check the market selector show the same markets as in the Compound web app. Order should be the same
3) Go market by market and check the Supply and Distribution APYs against the web app
4) General features from compound safe app should remain the same


mikheevm (approved):
perfecto

JagoFigueroa:
hey hey! testing latest changes from @yagopv 

Screenshot 2022-01-11 at 11 55 46

I think it would be interesting to specify both ' {{asset}} Supply APY' 'COMP Distribution APY' as no real feedback is provided of what those supply and distributions are (thinking of new users specially). Yago mentioned that it would be possible to add the token icons  which would match what Compound is showing to users:

Screenshot 2022-01-11 at 12 04 07

What do you think @dasanra ? Can perfectly be done for a future improvement ;)

JagoFigueroa:
Also, as discussed with @yagopv as well, there are a couple of tokens that come in this request (SAI & REP on mainnet) that are hidden in the compound app but we are showing  I don't know if we should do something for this cases @dasanra manually hide them maybe?

yagopv:
The contract for the markets come both from the API or if I read the [contract directly](https://etherscan.io/address/0x3d9819210a31b4961b30ef54be2aed79b9c9cd3b#readProxyContract) (see `getAllMarkets`)

So not sure what is the criteria for excluding them (perhaps manually)

PR #267, opened 2022-01-11
[TX Builder] Added valueError check to avoid adding tx with negative asset values

### LINKED ISSUE #263: [Tx Builder] Don't allow negative asset values
## Description
It is again possible to add negative asset values when using TX builder app when it shouldn't be

## Environment
- Environment: DEV (Prod is OK)

## Steps to reproduce
 1. Using Tx builder perform an Address to address tx and add a negative {{asset}} Value

## Expected result
Frontend validation pops up + unable to add transaction to the batch

## Obtained result
Although front validation pops up, it is possible to add transactions:

Screenshot 2022-01-10 at 15 09 34

Screenshot 2022-01-10 at 15 00 16

DaniSomoza:
## What it solves
Resolves #263 

## How this PR fixes it

Added valueError check in the addTransaction function to avoid adding transactions with negative asset values.

```
  const addTransaction = async () => {
    let description = '';
    let data = '';

    if (!services?.web3 || valueError) {
      return;
    }
```
## How to test it

Type a negative value in the asset value field and click on "Add Transaction" Button.

### DIFF
--- a/apps/tx-builder/src/components/Builder.tsx
+++ b/apps/tx-builder/src/components/Builder.tsx
@@ -208,7 +208,7 @@ export const Builder = ({
     let description = '';
     let data = '';
 
-    if (!services?.web3) {
+    if (!services?.web3 || valueError) {
       return;
     }
 


yagopv (approved):

JagoFigueroa:
Much bueno ;)

PR #273, opened 2022-01-13
[RAMP] Testing RAMP integration

yagopv:
## What it solves
Resolves https://github.com/gnosis/safe-apps-list/issues/95

## How this PR fixes it
This is a Safe APP POC using the [RAMP Widget](https://docs.ramp.network/configuration) for making Crypto purchases.

This Safe App is filtering coins by network so if you select Polygon [only MATIC can be purchased](https://docs.ramp.network/assets)

In the RAMP docs an [API key](https://docs.ramp.network/api-keys) is mentioned. Without the API key the widget works anyway (not tested in prod as it is a real purchase and a verification process is involved) but seems that only Revolut and Bank Account payments are allowed. For purchases with cards, you need [to be verified](https://discord.com/channels/607874487307993128/607881215671795712/885415706475261992) and have an Api key. If this app is really interesting for us we should talk to the RAMP team and verify all this info.

## How to test it
As this is a real purchase, we enabled test purchases in Rinkeby (minimum 0.05€). You can use Revolut. After the purchase you should receive the ETH in your safe

## Screenshots
image


mikheevm:
> For purchases with cards, you need to be verified and have an Api key. If this app is really interesting for us we should talk to the RAMP team and verify all this info.

I think we decided not to submit due diligence docs to them and go with unverified options. Just double-checking with @ennisjo

mikheevm:
does the solution without sdk allow card payments? https://ramp.network/buy/

yagopv:
By doing this you are losing prefilling the safe address and adding the coins supported in each network we are in

Apart from that, a most sophisticated app can be built using the widget in some element combined with the REST api for showing for example a list with purchase statuses

mikheevm:
> By doing this you are losing prefilling the safe address and adding the coins supported in each network we are in

Yes, but the last time I spoke to the Ecosystem team they said we weren't going to submit due diligence documents, therefore we won't add an API key. I'd ask the team if card payments are important

mikheevm:
**Important** Please, before releasing, let @lukasschor review the app to check if there are any legal concerns

apps/ramp-network/src/components/RampWidget.tsx
```diff
@@ -0,0 +1,43 @@
...
+        if (event.type === PURCHASE_CREATED_EVENT) {
+          // TODO: Send Analytics when the infra is ready
+          // https://github.com/gnosis/safe-apps-sdk/issues/255
+          console.log('PURCHASE_CREATED_EVENT', event);
+        }
+      })
+      .show();
+  }, [address, assets, onClose, url]);
+
+  return <div id="ramp-widget-container"></div>;
```
mikheevm, line 40: I'd move it to `index.html` because it always should be there in DOM, wdyt?

apps/ramp-network/src/components/RampWidget.tsx
```diff
@@ -0,0 +1,43 @@
...
+  const containerNode = useRef(null);
+
+  useEffect(() => {
+    new RampInstantSDK({
+      url,
+      hostAppName: 'Ramp Network Safe App',
+      hostLogoUrl: 'https://docs.ramp.network/img/logo-1.svg',
+      swapAsset: assets,
+      userAddress: address,
+      containerNode: containerNode.current || undefined,
```
mikheevm, line 24: we never set `containerNode`, is that correct?

apps/ramp-network/src/components/RampWidget.tsx
yagopv: mmm, I rewrote this because you are right. I can use a html component as the widget container but the way i was initialising it generates all the necessary HTML. So remove the components and created a initializeRampWidget function

apps/ramp-network/src/components/RampWidget.tsx
yagopv: yep

mikheevm (approved):
Me gusta el código

yagopv:
грасиас сеньор

JagoFigueroa:
I checked the app with @yagopv and it looks good as far as our testing goes. As commented in the daily, the problem comes when testing real transactions with real money for bank account and Revolut and payment methods. Putting this on hold for now.

dasanra (approved):

dasanra:
You can merge it so we can add it to dev and staging environments

yagopv:
Done

PR #288, opened 2022-01-17
[Drain Safe] Show transaction creation message

### LINKED ISSUE #281: [Drain Account] DEV - No tx creation message is shown
## Description
No tx creation message is shown to the user when a drain is performed.

## Steps to reproduce
Perform an account drain

## Expected result
This message is shown to the user to provide some feedback

Screenshot 2022-01-14 at 17 01 05

## Obtained result

No confirmation message is shown

yagopv:
## What it solves
Resolves #281 

## How this PR fixes it
We are adding the feedback message again and showing it for 5 seconds until automatically disappear

## How to test it
1) Open Drain Safe
2) Send transaction 
3) The message should be showed and kept on screen for 5 seconds

### DIFF
--- a/apps/drain-safe/src/hooks/useTimeout.ts
+++ b/apps/drain-safe/src/hooks/useTimeout.ts
@@ -1,0 +1,21 @@
+import { useEffect, useLayoutEffect, useRef } from 'react';
+
+function useTimeout(callback: () => void, delay: number | null) {
+  const savedCallback = useRef(callback);
+
+  useLayoutEffect(() => {
+    savedCallback.current = callback;
+  }, [callback]);
+
+  useEffect(() => {
+    if (!delay && delay !== 0) {
+      return;
+    }
+
+    const id = setTimeout(() => savedCallback.current(), delay);
+
+    return () => clearTimeout(id);
+  }, [delay]);
+}
+
+export default useTimeout;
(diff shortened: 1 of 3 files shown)


apps/drain-safe/src/hooks/useTimeout.ts
```diff
@@ -0,0 +1,21 @@
+import { useEffect, useLayoutEffect, useRef } from 'react';
+
+function useTimeout(callback: () => void, delay: number | null) {
```
DaniSomoza, line 3: What is your opinion to avoid useRef here and call directly to the callback if no delay is provided?

```
import { useEffect } from 'react';

function useTimeout(callback: () => void, delay: number | null) {
  useEffect(() => {
    if (!delay && delay !== 0) {
      callback();
      return;
    }

    const id = setTimeout(() => callback(), delay);

    return () => clearTimeout(id);
  }, [delay, callback]);
}

export default useTimeout;

```

yagopv: yep, that's because the outside callback can change because another rerender and with the ref we keep track of the original one

DaniSomoza (approved):
Approved!

JagoFigueroa:
All bueno ;)

PR #289, opened 2022-01-17
[Drain Safe] Fix button message

### LINKED ISSUE #284: [Drain Account] native token still displayed as asset to be drained when value = 0 (Button only)
## Description
The native token is still counted as an asset to be drained when it is hidden in the drain account table (this happens in dev only & only the button displays it wrong, no extra tx is added or anything)

## Steps to reproduce
Mock a drain of all assets in: https://safe-team.dev.gnosisdev.com/app/rin:0x70F2c8773c5EBbCD3a9a05B67491EF41688a4d8f/apps?appUrl=https%3A%2F%2Fsafe-apps.dev.gnosisdev.com%2Fdrain-safe

## Expected result
For safe above, button should say: Transfer Everything, everything being wETH only

## Obtained result
User is given feedback as 2 assets are going to be drained when it is not true.

Screenshot 2022-01-17 at 10 32 37

yagopv:
## What it solves
Resolves #284 

## How this PR fixes it
By calculating from the same filtered assets both the selected and the ones showed in the table

## How to test it
1) Open Drain Safe with an account with a native token with value 0 among other assets
2) Check the display message in the button is showing the correct number of assets

### DIFF
--- a/apps/drain-safe/src/hooks/use-balances.ts
+++ b/apps/drain-safe/src/hooks/use-balances.ts
@@ -26,8 +26,10 @@ function useBalances(safeAddress: string, chainId: number): BalancesType {
 
     try {
       const balances = await sdk.safe.experimental_getBalances({ currency: 'USD' });
-      setAssets(balances.items.filter(transferableTokens));
-      setSelectedTokens(balances.items.map((token: any) => token.tokenInfo.address));
+      const assets = balances.items.filter(transferableTokens);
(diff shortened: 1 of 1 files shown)


dasanra (approved):

DaniSomoza (approved):
Approved! :D

JagoFigueroa:
All bueno again ;)

Issue #219, opened 2021-11-29
[Safe Apps] Create automatic changeset and publish


dasanra:
## Overview
Currently it's difficult to keep track of the changes on apps. We should create an automated changelog, something similar to what we have on safe-apps-sdk: [Example PR](https://github.com/gnosis/safe-apps-sdk/pull/259)

## Goals
Add automation to changeset creation to keep track and version the different safe apps when changes are added.
Automatically build the new app versions to be deployed on prod

PR #302, opened 2022-01-25
build: add lerna and individual deploy commands to each package

mikheevm:
## What it solves
This PR sets a base for #219 by:
- Adding `lerna` to the monorepo
- Adding `commitizen` for conventional commits (please install it globally on your computer). You can read about commits [here](https://www.conventionalcommits.org/en/).
- Adjust deployment scripts to run a deployment for one app and add deploy script to each app
- Use lerna in GitHub actions

## How to test it
- CI/CD should work
- running `git commit` on your computer should open a CLI tool for commit messages


mikheevm:
I will close this because today we learned that lerna isn't maintained anymore. I'll explore other options

PR #314, opened 2022-02-08
[TX Builder] Separate address field in address and ABI

### LINKED ISSUE #305: [TX-Builder] Use separated inputs for ABI and address
## Overview
Currently the UX with unified field in TX-Builder for ABI and address can be quite confusing. Some users told us that they feel confused wether it makes sense to set an ABI or an address and in which cases one would make sense over the other one. Also once the user inputs an address we show a second address input where most of the times the same address has to be added there

## Goals
Create a better UX, separating ABI and address inputs. The address input will be also used as the `to` address. When adding an address if the ABI field is empty we will check if we can automatically fetch the ABI.

Before overwritting an existing ABI we should ask the user if they want to replace the ABI

## Screens
 - Figma:
https://www.figma.com/proto/G7snLqBXWNPANcOj5MhiR7/Transaction-Builder?page-id=1%3A2&node-id=55%3A17805&viewport=241%2C48%2C0.12&scaling=min-zoom&starting-point-node-id=55%3A16961

yagopv:
## What it solves
Resolves #305 

## How this PR fixes it
We are adding a new JSON Field for displaying and managing the Contract ABI interaction. Now the address and the ABI are separated.

## How to test it
1) Enter valid address with retrieved ABI. JSON field should be filled in after the correct retrieval and the methods interaction showed up
2) Enter valid address with no ABI retrieval should show only the ETH send 
3) You can edit the ABI manually
4) If you try to paste in the ABI field when it has some value a popup should be showed asking if the ABI replacement is intended

**TX Builder should work same way adding and sending transactions**

## Screenshots

image

image

image


apps/tx-builder/src/components/Dashboard.tsx
```diff
@@ -108,39 +115,34 @@ const Dashboard = (): ReactElement => {
...
         }}
       />
 
-      {/* ABI Warning */}
-      {isValidAddressOrContract && !contract && (
-        <StyledWarningText color="warning" size="lg">
-          No ABI found for this address
-        </StyledWarningText>
-      )}
+      <JsonField id={'abi'} name="abi" label="Enter JSON" value={abi} onChange={setAbi} />
```
DaniSomoza, line 138: I think that we can use another label right? because I think that "Enter JSON" is too generic for the user
Something like "Enter ABI" ?

Captura de pantalla 2022-02-09 a las 10 54 05

apps/tx-builder/src/components/Dashboard.tsx
```diff
@@ -108,39 +115,34 @@ const Dashboard = (): ReactElement => {
...
         }}
       />
 
-      {/* ABI Warning */}
-      {isValidAddressOrContract && !contract && (
-        <StyledWarningText color="warning" size="lg">
-          No ABI found for this address
-        </StyledWarningText>
-      )}
+      <JsonField id={'abi'} name="abi" label="Enter ABI" value={abi} onChange={setAbi} />
```
DaniSomoza, line 152: We are not showing errors in the JsonField. Do you think that we can add it?
Captura de pantalla 2022-02-09 a las 11 19 07

As a suggestion something like:

```
function isValidJSON(value: string | undefined) {
  if (value) {
    try {
      JSON.parse(value);
    } catch {
      return false;
    }
  }

  return true;
}
```

```
      <JsonField
        id={'abi'}
        name="abi"
        label="Enter ABI"
        value={abi}
        onChange={setAbi}
        error={isValidJSON(abi) ? '' : 'Invalid JSON value'}
      />
```
Captura de pantalla 2022-02-09 a las 11 31 23

apps/tx-builder/src/components/Dashboard.tsx
```diff
@@ -53,41 +54,54 @@ const StyledAddressInput = styled(AddressInput)`
...
+
+    setContract(interfaceRepo.getMethods(abi));
+  }, [abi, interfaceRepo]);
 
   const getAddressFromDomain = (name: string): Promise<string> => {
     return web3?.eth.ens.getAddress(name) || new Promise((resolve) => resolve(name));
   };
 
-  const isValidAddressOrContract = (isValidAddress(addressOrAbi) || contract) && !isABILoading;
+  const contractHasMethods = contract && contract.methods.length > 0 && !isABILoading;
```
DaniSomoza, line 101: I am not 100% sure about adding this extra condition about contract methods length here...

if you use a contract ABI without methods:
```
[
            {
              "inputs": [],
              "stateMutability": "nonpayable",
              "type": "constructor"
            }

]
```
we are not showing the form:

Captura de pantalla 2022-02-09 a las 11 50 02

yagopv: aha. No public methods available. Don't know, the current UI shows a message and the address field but seems as an bug because you cannot do anything with that

yagopv: perhaps we can show a warning

DaniSomoza: The current UI shows only the "To Address" field (And the possibility to use a custom hex data encoded) 

Captura de pantalla 2022-02-09 a las 11 56 42

I think that we can go ahead with this new condition.

apps/tx-builder/src/components/forms/fields/JsonField.tsx
```diff
@@ -0,0 +1,171 @@
...
+        toggleModal();
+      } else {
+        onChange(pastedData);
+      }
+    },
+    [onChange, toggleModal, value],
+  );
+
+  return (
+    <>
```
mikheevm, line 63: is `<>` really needed here?

apps/tx-builder/src/components/forms/fields/JsonField.tsx
```diff
@@ -0,0 +1,171 @@
...
+  const [tempAbi, setTempAbi] = useState(value);
+  const [isPrettified, setIsPrettified] = useState(false);
+
+  const toggleFormatJSON = useCallback(() => {
+    if (!value) {
+      return;
+    }
+
+    try {
+      onChange?.(JSON.stringify(JSON.parse(value), null, isPrettified ? 0 : 2));
```
mikheevm, line 28: The type for onChange doesn't allow undefined, why do we do `?` here? It's the compiler's job to catch if we pass `undefined` here

apps/tx-builder/src/components/Dashboard.tsx
yagopv: Updated this. Nice!

apps/tx-builder/src/components/forms/fields/JsonField.tsx
yagopv: yep because I have a Modal under the JSON Field

apps/tx-builder/src/components/forms/fields/JsonField.tsx
yagopv: yeah you are right. I think this remained from a previous implementation i tried

mikheevm (approved):
Vamos

DaniSomoza (approved):
Great work bro! 💪

JagoFigueroa:
All bueno, thank you ;)

PR #335, opened 2022-02-17
feat(wallet-connect): Check if app exists as a Safe App

### LINKED ISSUE #167: [WalletConnect] Show message when user have connected to an Dapp that exists as a Safe App
## Overview
We know that users continue using Dapps via wallet connect when there is a Safe App that exists. Why? 
We are not 100% sure but from observing the Defi team we know that once an Tx is done, they get used doing it in a specific way, i.e with Wallet Connect. 

Why is this a problem? 

Mainly because of wallet connect reliability. This stems from projects using the default demo bridge, which is not always up. 

## Goals
As a user of the safe wallet connect app,
when I connect to a Dapp that exists as a Safe App,
I should be informed that it exists as a Safe App,
so that I know, in case of issues, I can always use the Safe App. 

concept design - Not for production:
image

## Dependencies
- [ ] We need to state a method to get the list of Safe Apps. We should decide whether to add it to the SDK or do the direct call to the backend
- [ ] Many apps still use a custom version hosted on IPFS so we can't do a domain match in order to detect that the app used through wallet connect is an app that already have a proper Safe App. The solution could be to add a new value in the config service setting the root domain of the app where the custom safe app is based - create ticket for backend services team.

yagopv:
## What it solves
resolves #167 

## How this PR fixes it
We are checking if an app exists as a Safe App. In this case we are asking the user for decide to use the Safe App or continue using Wallet Connect

As part of this task we are revamping the whole UI following this designs and workflow

https://www.figma.com/file/iicHRaXd0ypyttJqq2WMga/Wallet-Connect?node-id=13%3A3195

## How to test it
1) Open Wallet Connect app
2) Open any Dapp as Uniswap o Sushiswap in a new window and try to connect
3) When connecting => If the app exists as a safe app should prompt the user to open it if preferred
4) User can choose to continue using the Wallet Connect app
5) When executing transaction from the dapps the safe popup should appear

Check figma designs for the complete flow


apps/wallet-connect/package.json
```diff
@@ -5,6 +5,7 @@
   "homepage": "./",
   "dependencies": {
     "@gnosis.pm/safe-react-components": "^0.9.7",
+   "@gnosis.pm/safe-react-gateway-sdk": "^2.8.5",
```
mikheevm, line 8: Did we sync with the infra team if it's ok to use the gateway? so far, the gateway was meant to be used in clients only

apps/wallet-connect/src/assets/qr-code.svg
```diff
@@ -0,0 +1,14 @@
+<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
dasanra, line 1: https://components.gnosis-safe.io/?path=/docs/data-display-icon--icons

There is already `qrCode` here

apps/wallet-connect/src/assets/safe-app-connect.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="44" height="44" viewBox="0 0 44 44" fill="none" xmlns="http://www.w3.org/2000/svg">
```
dasanra, line 1: https://components.gnosis-safe.io/?path=/docs/data-display-icon--icons

There is already `apps` here

apps/wallet-connect/package.json
dasanra: We synced, is expected that in the future will be a gate for external devs to fetch info directly from backend. Only limitation they warned us is that client-gateway is meant for Safe clients and thereby backward compatibility is not guaranteed, endpoints may break in future versions. 

As we are integrating with the same package needed for web we would be fine in terms of adapting to gateway changes and both teams can help each other to keep it updated

apps/wallet-connect/src/assets/qr-code.svg
yagopv: yeah, that's the one currently in prod but the @liliiaorlenko one is different, for this reason I imported it here as SVG

yagopv: Should we use the library one anyway?

apps/wallet-connect/src/assets/safe-app-connect.svg
yagopv: yep, that's right. Will use that

DaniSomoza (approved):
Awesome Job! :D

Looks great! :3

JagoFigueroa:
Hola guys! Loving the new design. I have a couple of suggestions.

For this screen:
Screenshot 2022-02-22 at 16 13 16

When a user reaches it connection has already been stablished via wallet connect. What we do is:

-if the user chooses to open the dapp we terminate the session and open the safe app.

-If the user chooses to keep using wallet connect we keep it. 

Maybe we should change the trying to connect message?

Regarding that screen also, if we refresh the page the session is lost (which is expected) but if we instead leave the app and move to another section on the safe the session will be kept. I think it would be to terminate it as well.

What do you think @dasanra @yagopv @liliiaorlenko ?

liliiaorlenko:
> Hola guys! Loving the new design. I have a couple of suggestions.
> 
> For this screen: Screenshot 2022-02-22 at 16 13 16
> 
> When a user reaches it connection has already been stablished via wallet connect. What we do is:
> 
> -if the user chooses to open the dapp we terminate the session and open the safe app.
> 
> -If the user chooses to keep using wallet connect we keep it.
> 
> Maybe we should change the trying to connect message?
> 
> Regarding that screen also, if we refresh the page the session is lost (which is expected) but if we instead leave the app and move to another section on the safe the session will be kept. I think it would be nice to terminate it as well.
> 
> What do you think @dasanra @yagopv @liliiaorlenko ?

I think better to keep message as is it, because no matter if we already connected app or not, we need to encourage user to use Safe app instead and creating passive message that "you have already been connected but if u want u can user Safe app" is not ideal strategy. In this case not showing exact state of the system is better for us to not let person use app that is optimised :P 

Also, do you think we need a designs for the screen when we show that the last connection has been lost, in case user made a mistake? @dasanra 

And another question from the design: Are we changing the style of the buttons or its more depended on a react team?

yagopv:
`Also, do you think we need a designs for the screen when we show that the last connection has been lost, in case user made a mistake?` => Not sure about what's this about :( When we show that state ?

`Are we changing the style of the buttons or its more depended on a react team?` => Do you mean about the button not being exactly the same as the one in the design ? We are using a library under our control => https://components.gnosis-safe.io/?path=/docs/inputs-button--simple-button

JagoFigueroa:
All bueno, gracias Yago ;)

PR #342, opened 2022-02-21
fix(compound): App failing if token liquidity retrieved from REST API is zero

### LINKED ISSUE #341: [Compound app] Prod mainnet app not working + staging issue
## Description
Prod & staging safe app is not loading for mainnet.

## Steps to reproduce
Open mainnet safe app.

## Expected result
App opens & working as expected.

## Obtained result
Infinite loader shown for staging and prod mainnet apps.

Rinkeby one is loading but on staging no asset values are fetched.

Edit: after some research, the problem is that Compound added a new token on their API with 0 value, without decimals. As we are always expecting decimals, the app is not loading correctly.

yagopv:
## What it solves
Resolves #341 

## How this PR fixes it
We are filtering out the coins with no liquidity


DaniSomoza (approved):
Approved! :D

dasanra (approved):

JagoFigueroa:
App is not crashing anymore + working nicely on mainnet. Thanks for the quick action ;)

Issue #316, opened 2022-02-08
[Tx Builder] Booleans always set to true when using testArrayBool method


JagoFigueroa:
## Description
Booleans are always recognised as **true** when using this method on Tx Builder

## Steps to reproduce
 1. Go to tx builder on RInkeby and use this testing smart contract: `0x49d4450977E2c95362C13D3a31a09311E0Ea26A6`
 2. Choose testArrayBool method from the list
 3. add a tx to the batch as `["true"]` and another as `["false"]`
 4. Check tx details

## Expected result
1 tx set as true and the other as false

## Process + obtained result

https://user-images.githubusercontent.com/92332710/153019964-636553a0-10dd-4f28-99cd-7b1a38f6a43f.mov

PR #350, opened 2022-02-25
fix(tx-builder): parse testBoolArray values in a different way as the encode method don't understand strings

yagopv:
## What it solves
Resolves https://github.com/gnosis/safe-react-apps/issues/316

## How this PR fixes it
When encoding bool[] type, if the array contains strings as "true"or "false" the encode result is always true values. We had same issue for normal bool type. 

For resolving this we are going to parse the array en convert the values to proper booleans


DaniSomoza (approved):
Approved! :D

dasanra (approved):

Issue #348, opened 2022-02-23
[TX Builder] Loading transaction batch from a file


dasanra:
## Overview
We want users to be able to store and edit a batch locally, not only as a convenient way of repeating a batch that they need to execute often, but also as a way of share with other teams and parties that may be interested on executing the exact same operations. Also for really advanced user it could be a way of creating transactions locally and doing only one step at time of adding it to the transaction builder, as scripts could be used to generate such kind of structured data.

## Goals
We should enable the users a way to export the batch they create and to load it again with low friction. Also in case the structure is correctly edited, advanced users would be able to create a script that returns as outcome a transaction batch file that the transaction builder should be able to handle

## Requirements
 - Create a standard structured file to import/export transactions. Is really important to first find an agreement on the structure of the file, as won't be as easy to migrate as localstorage that is not that easy to manipulate by the user.
 - When adding a file the batch should directly be loaded
 - When exporting as file the downloaded file should follow the same semantic as we expect the user to upload

## Screens
https://www.figma.com/file/G7snLqBXWNPANcOj5MhiR7/Transaction-Builder?node-id=828%3A7471

## Links
Related problem statement https://github.com/gnosis/safe-pm/issues/38

PR #352, opened 2022-02-28
feat(tx-builder)!: transaction library feature

### LINKED ISSUE #349: [TX Builder] Transaction library
## Overview
We want users to be able to store a batch in a localstorage library, not only as a convenient way of repeating a batch that they need to execute often, but also as a way of always having it available on their device.

## Goals
We should enable the users a convenient way of storing transactions in their browser in a similar way they are able to handle the address book.

## Requirements
 - User is able to save the batch in the transaction library
 - User is able to load a batch from the transaction library
 - User is able to export a batch from the transaction library
 - User is able to edit a batch from the transaction library

## Screens
https://www.figma.com/file/G7snLqBXWNPANcOj5MhiR7/Transaction-Builder?node-id=828%3A7471

## Links
Related problem statement https://github.com/gnosis/safe-pm/issues/38

DaniSomoza:
## What it solves
Resolves #349 & #348 

Figma: https://www.figma.com/file/G7snLqBXWNPANcOj5MhiR7/Transaction-Builder?node-id=939%3A7718

## Screenshots
Captura de pantalla 2022-03-10 a las 14 43 00
Captura de pantalla 2022-03-10 a las 14 43 13
Captura de pantalla 2022-03-10 a las 14 43 19
Captura de pantalla 2022-03-10 a las 14 43 36
Captura de pantalla 2022-03-10 a las 14 44 19
Captura de pantalla 2022-03-10 a las 14 44 53
Captura de pantalla 2022-03-10 a las 14 45 09
Captura de pantalla 2022-03-10 a las 14 45 50


apps/tx-builder/src/components/ChecksumWarning.tsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import MuiAlert from '@material-ui/lab/Alert';
+import MuiAlertTitle from '@material-ui/lab/AlertTitle';
+import styled from 'styled-components';
+import { useTransactionLibrary } from '../store';
+
```
DaniSomoza, line 6: Captura de pantalla 2022-03-10 a las 14 57 55

apps/tx-builder/src/components/CreateNewBatchCard.tsx
```diff
@@ -1,15 +1,94 @@
...
 
 import { ReactComponent as CreateNewBatchSVG } from '../assets/add-new-batch.svg';
+import useDropZone from '../hooks/useDropZone';
 
-const CreateNewBatchCard = () => {
-  return <StyledArrowToFormSVG />;
+type CreateNewBatchCardProps = {
+  onFileSelected: (file: File | null) => void;
+};
+
```
DaniSomoza, line 12: Captura de pantalla 2022-03-10 a las 14 58 43

apps/tx-builder/src/components/QuickTip.tsx
```diff
@@ -0,0 +1,50 @@
+import { Icon } from '@gnosis.pm/safe-react-components';
+import MuiAlert from '@material-ui/lab/Alert';
+import MuiAlertTitle from '@material-ui/lab/AlertTitle';
+import React from 'react';
+import styled from 'styled-components';
+
+type QuickTipProps = {
+  onClose: () => void;
+};
+
```
DaniSomoza, line 10: Captura de pantalla 2022-03-10 a las 15 00 14

apps/tx-builder/src/components/modals/DeleteBatchFromLibrary.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import styled from 'styled-components';
+import { Batch } from '../../typings/models';
+
+type DeleteBatchFromLibraryProps = {
+  batch: Batch;
+  onClick: (batch: Batch) => void;
+  onClose: () => void;
+};
+
```
DaniSomoza, line 12: Captura de pantalla 2022-03-10 a las 15 02 45

apps/tx-builder/src/components/modals/SaveBatchModal.tsx
```diff
@@ -0,0 +1,83 @@
...
+type SaveBatchModalProps = {
+  onClick: (name: string) => void;
+  onClose: () => void;
+};
+const BATCH_NAME_FIELD = 'batchName';
+
+type CreateBatchFormValuesTypes = {
+  [BATCH_NAME_FIELD]: string;
+};
+
```
DaniSomoza, line 22: Captura de pantalla 2022-03-10 a las 15 03 23

apps/tx-builder/src/components/modals/downloadBatchModal.tsx
```diff
@@ -0,0 +1,64 @@
...
+type SaveBatchModalProps = {
+  onClick: (name: string) => void;
+  onClose: () => void;
+};
+const BATCH_FILE_NAME_FIELD = 'batchFileName';
+
+type CreateBatchFormValuesTypes = {
+  [BATCH_FILE_NAME_FIELD]: string;
+};
+
```
DaniSomoza, line 18: Captura de pantalla 2022-03-10 a las 15 04 37

apps/tx-builder/src/pages/CreateTransactions.tsx
```diff
@@ -0,0 +1,115 @@
...
+import Grid from '@material-ui/core/Grid';
+
+import TransactionsBatchList from '../components/TransactionsBatchList';
+import CreateNewBatchCard from '../components/CreateNewBatchCard';
+import { CREATE_BATCH_PATH, REVIEW_AND_CONFIRM_PATH } from '../routes/routes';
+import QuickTip from '../components/QuickTip';
+import { useNetwork, useTransactionLibrary, useTransactions } from '../store';
+import useModal from '../hooks/useModal/useModal';
+import WrongChainBatchModal from '../components/modals/WrongChainBatchModal';
+
```
DaniSomoza, line 15: Captura de pantalla 2022-03-10 a las 15 08 14

apps/tx-builder/src/pages/SaveTransactionLibrary.tsx
```diff
@@ -0,0 +1,66 @@
+import { useEffect } from 'react';
+import { useNavigate } from 'react-router-dom';
+import Grid from '@material-ui/core/Grid';
+import styled from 'styled-components';
+import { Button } from '@gnosis.pm/safe-react-components';
+
+import TransactionsBatchList from '../components/TransactionsBatchList';
+import { useTransactionLibrary, useTransactions } from '../store';
+import { CREATE_BATCH_PATH, TRANSACTION_LIBRARY_PATH } from '../routes/routes';
+
```
DaniSomoza, line 10: Captura de pantalla 2022-03-10 a las 15 11 19

package.json
```diff
@@ -23,6 +23,7 @@
   "dependencies": {
     "@gnosis.pm/safe-apps-react-sdk": "4.2.0",
     "@gnosis.pm/safe-apps-sdk": "6.3.0",
+    "localforage": "^1.10.0",
```
mikheevm, line 26: this should be put into the tx-builder package.json because it's the only app using this dependency

apps/tx-builder/src/components/TransactionBatchListItem.tsx
```diff
@@ -0,0 +1,352 @@
...
+              </StyledAccordion>
+            </TransactionListItem>
+          );
+        }}
+      </Draggable>
+    );
+  },
+);
+
+export const getTransactionText = (description: ProposedTransaction['description']) => {
```
mikheevm, line 197: would move it to a file with utils, the component is already big enough without additional logic, plus it is exported and used in other components

apps/tx-builder/src/components/modals/WrongChainBatchModal.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+const WrongChainBatchModal = ({ onClick, onClose }: WrongChainBatchModalProps) => {
+  return (
+    <GenericModal
+      title="Warning!"
+      withoutBodyPadding
+      body={
+        <StyledModalBodyWrapper>
+          <Text size={'xl'} center>
+            This Batch is from another Chain!
```
mikheevm, line 18: I'd change the working to something like, "This batch was created on a different network. Please proceed with caution and make sure the contracts exist on the current network". You can also ask a native speaker :D

apps/tx-builder/src/lib/checksum.tsx
```diff
@@ -0,0 +1,48 @@
+import SHA256 from 'crypto-js/sha256';
```
mikheevm, line 1: Why couldn't we use sha3 from ethers or web3?

mikheevm (approved):

apps/tx-builder/src/lib/storage.tsx
```diff
@@ -0,0 +1,108 @@
+import localforage from 'localforage';
```
mikheevm, line 1: this file doesn't use jsx, so `.tsx` file extension is not needed

apps/tx-builder/src/lib/checksum.tsx
mikheevm: this file doesn't use jsx, so `.tsx` file extension is not needed

apps/tx-builder/src/lib/checksum.test.js
```diff
@@ -0,0 +1,103 @@
+import { addChecksum, validateChecksum } from './checksum';
```
mikheevm, line 1: Why are tests written in javascript and not typescript?

mikheevm: How is the `lib` folder different from the `services` folder? I'd avoid spreading logic code across different folders

apps/tx-builder/src/components/Header.tsx
```diff
@@ -1,35 +1,69 @@
...
+  const { batches } = useTransactionLibrary();
+
+  const isTransactionCreationPath = pathname === CREATE_BATCH_PATH;
+  const isSaveBatchPath = pathname === SAVE_BATCH_PATH;
+
+  const showTitle = isTransactionCreationPath || isSaveBatchPath;
+  const showLinkToLibrary = isTransactionCreationPath || isSaveBatchPath;
+
+  const { state } = useLocation();
+  const { from: previousUrl } = (state as { from: string }) || { from: CREATE_BATCH_PATH };
```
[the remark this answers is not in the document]
DaniSomoza: I simplified it using:
```
  const { state } = useLocation() as LocationType;

  const previousUrl = state?.from || CREATE_BATCH_PATH;
```

apps/tx-builder/src/components/forms/AddNewTransactionForm.tsx
```diff
@@ -1,44 +1,38 @@
...
+import { ContractInterface } from '../../typings/models';
 import { isValidAddress } from '../../utils';
 import SolidityForm, {
   CONTRACT_METHOD_INDEX_FIELD_NAME,
   SolidityFormValuesTypes,
   TO_ADDRESS_FIELD_NAME,
   parseFormToProposedTransaction,
 } from './SolidityForm';
+import { useTransactions } from '../../store';
+import { useNetwork } from '../../store';
```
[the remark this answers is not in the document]
DaniSomoza: done!

package.json
yagopv: true, totally unintentional :(

apps/tx-builder/src/components/TransactionBatchListItem.tsx
yagopv: Good idea

apps/tx-builder/src/lib/checksum.test.js
yagopv: I use this test file for helping me to  develop the checksum feature and left it in the code. Is the only test file in the project right now and didn't see the benefit of converting it to TS for this concrete case to be honest

yagopv: going to move all to the lib folder!

apps/tx-builder/src/lib/checksum.tsx
mikheevm: yes!

JagoFigueroa:
Hola! starting my testing for this padrísima feature. 

There is a small issue that allows to edit a batch name to the point of removing it. This is causing a couple of issues. 

Without a name the save / download / delete component goes a bit crazy and moves to the left
componente sin batch name

And downloading a file without a name does not download it in JSON format. 
Screenshot 2022-03-17 at 16 32 13

Maybe we should make the name required on the edit batch screen?

Cheers!

JagoFigueroa:
hola again :pepe-evil-qa:

- What about providing some stronger feedback when the uploaded batch is invalid? this warning looks like it was meant to pop up for a modified but still valid and re-uploaded batch.

Screenshot 2022-03-17 at 16 58 25 1

- Should we trim the downloaded file names to avoid issues caused by spacing?
Screenshot 2022-03-17 at 17 11 53


- (not related to this task but yolo) Should we make the app usable on smaller browser window sizes? at the moment the batch component is hidden completely instead of being visible under the new transaction one:

Screenshot 2022-03-17 at 17 06 28

JagoFigueroa:
Morning! a few more suggestions from my side:

- Drag and drop goes a bit crazy specially when there are enough transactions on the batch for the scrolling bar to pop up. Uploading a video as a quick demo:

https://user-images.githubusercontent.com/92332710/158990618-cddae41f-adf3-474b-ade4-77a7271983c7.mov

- I think we should allow to edit the batch name on the edit batch screen as well, not only from the batch library view.

- Once the save to library button becomes green because we have saved a batch it will keep that color for other potential batches to save on the library. Quick example:

https://user-images.githubusercontent.com/92332710/158993252-ea8ed781-d728-40ee-8715-24b533c1115d.mov

- Navigation is a bit buggy on edit batch screen. See that after editing one, the 'back to Tx creation' is redirecting to edit batch instead: 

https://user-images.githubusercontent.com/92332710/158993883-201f3b46-7472-46bc-a963-c780f09de59e.mov

-Saving a batch is redirecting the user to the save batch screen but the header is not updated to reflect this, making hard to navigate to other screens.


https://user-images.githubusercontent.com/92332710/158994408-673dbb9f-19f8-46c0-bc58-d18bd2bdaffb.mov

JagoFigueroa:
A nice catch by @DaniSomoza naming the batch can get a bit crazy:

Screenshot 2022-03-18 at 13 15 33

Maybe we should limit the name length ;)

yagopv:
@JagoFigueroa I addressed some of the virtualization issues. Can you take a look?

JagoFigueroa:
Only because it's you @yagopv ;)

yagopv:
> What do you think (or is it possible & worth it) to ignore the batch name from the checksum in order to reduce the notifications we show to the users?

Yep, make sense

yagopv:
> small issue, when the batch name is too long on the edit batch screen it affects the number of txs indicator

Pushed fix

yagopv:
> What do you think (or is it possible & worth it) to ignore the batch name from the checksum in order to reduce the notifications we show to the users?

Fixed!

JagoFigueroa:
Question. Lets say I have a batch that I created from a smart contract. When I am creating this batch for the first time, I will be able to edit transactions and switch between different methods if needed.

When I edit the batch itself, that method switching on each transaction of the batch is not possible anymore which makes sense as that smart contract info is no longer there to be used.

But, if I go back to the tx creation screen and use the smart contract address again for adding more transactions to the batch. Should I be able to switch between methods for the saved batch / transaction again?

https://user-images.githubusercontent.com/92332710/159505643-ea9d69ab-d795-4e10-84fd-97757393002a.mov

I feel that being able to switch between methods for a tx relies too heavily on only being able to do it once, when the batch is created for the first time. Then the method selector becomes just a selector with one option:

Screenshot 2022-03-22 at 15 32 16

JagoFigueroa:
All bueno from my side after the latest changes ;)

PR #381, opened 2022-03-14
fix: missing infura token 

dasanra:
## What it solves
Solves an issue where some queries to Infura couldn't be resolved

## How this PR fixes it
Restores the Infura RPC API configuration in the build steps that require it

## How to test it
 - Test that Compound can fetch balances on mainnet


mikheevm (approved):
oops

Issue #359, opened 2022-03-07
[Tx Builder] EOA address not clearing ABI field


JagoFigueroa:
## Description
If after a smart contract address / ABI itself populates the ABI field we input an EOA address or ENS, the ABI field will not be cleared.

## Steps to reproduce
1. Copy / Paste an address with an ABI or an ABI directly.
2. Once populated Enter a EOA address or ENS

https://user-images.githubusercontent.com/92332710/157034888-e3709bbc-5e23-4515-9153-f8454c677db5.mov

## Expected result
EOA address should also clear ABI field

## Obtained result
EOA address not clearing ABI field

PR #386, opened 2022-03-15
fix(tx-builder): JsonField value when tx service abi returns undefined

yagopv:
## What it solves
Resolves https://github.com/gnosis/safe-react-apps/issues/359

## How this PR fixes it
By setting an empty string in case the return value from tx-service is null/undefined


dasanra (approved):

JagoFigueroa:
fixed! thank you for your magic Yago ;)

Issue #313, opened 2022-02-08
[Sushi app] app not recognised on tx submit pop up 


JagoFigueroa:
## Description
App logo and name are not fetched correctly when on submit tx pop up.

## Steps to reproduce
Go to sushi app and mock a tx without sending it.

(Was able to reproduce this on on Gnosis Chain + Mainnet)

## Expected result
App info is displayed on pop up.

## Obtained result
![Screenshot 2022-02-08 at 09 03 47](https://user-images.githubusercontent.com/92332710/152946124-f69c3e24-fbda-4e90-a64d-754f191e1f46.png)
+
![Screenshot 2022-02-08 at 09 16 35](https://user-images.githubusercontent.com/92332710/152945832-01a386d7-7780-4964-9ee3-891f322455b9.png)

yagopv:
The Sushi app [manifest](https://app.sushi.com/manifest.json)  does not comply our rules and the `description` and `iconPath` are missing so this is in the sushi app roof :( @JagoFigueroa @dasanra

dasanra:
I think we can use this as oportunity to follow manifest.json specification more tightly.
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json

Description we should allow to be optional.

Also we should transition `iconPath` to `icons` structure, expecting an array and some way of fetching some of the icons included there, although people is not following 100% the standard there. Icons is also optional.

In those cases where we can't find this information we can always try to fallback to the backend information. If it wasn't a manually added app the info should be available there

yagopv:
I can do it. So should be:

1) Make `description` optional
2) Try to find image
  2.1) `icons`
  2.2) when not `icons` try `iconPath`
  2.3) If not `icons` and no `iconPath` then fallback to backend

JagoFigueroa:
Looks good to me, thank you guys ;)

PR #388, opened 2022-03-16
feat: Standarize manifests

yagopv:
# What it solves
resolves https://github.com/gnosis/safe-react-apps/issues/313

# How this PR fixes it


apps/wallet-connect/public/manifest.json
```diff
@@ -1,8 +1,11 @@
...
-  "short_name": "WalletConnect",
   "name": "WalletConnect",
   "description": "Allows your Gnosis Safe Multisig to connect to dapps via WalletConnect.",
-  "iconPath": "wallet-connect.svg",
-  "theme_color": "#000000",
-  "background_color": "#ffffff"
+  "icons": [
+    {
+      "src": "wallet-connect.svg",
+      "sizes": "any",
```
mikheevm, line 7: is `any` a part of standard?

yagopv: I extracted it from here:

- https://developer.mozilla.org/en-US/docs/Web/Manifest/icons (Link to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-sizes)

So I think so, but taking advantage of your comment I'm going to look up as well for `any` in the `purpose` prop

mikheevm: Awesome. Also wanted to share that the spec from w3c sometimes contradicts MDN: https://www.w3.org/TR/appmanifest/

I remember this from my adventures with manifests in the past 😄

mikheevm (approved):
buen trabajo amigo

apps/wallet-connect/public/manifest.json
yagopv: Yeah I looked as well to the spec and cannot see anything related to the any value. But still I think is really used

dasanra (approved):

PR #391, opened 2022-03-21
feat(tx-builder): tenderly simulation

### LINKED ISSUE #200: [Tx Builder] Research how to simulate state change of proposed batch transaction
## Overview
Executing a huge batch transaction can get really complicated. Once the user spent plenty of time creating the batch, they may find that executing that batch "could fail", because this is the only feedback we are giving to our users. To reduce the chance of creating a wrong transaction batch we need to give some early feedback for our users. This is specially important when using the transaction builder. 

## Goals
Currently there isn't a clear approach on how to tackle this problem. There are different levels of solutions:
 - Pointing up to witch transaction the batch is failing
 - Pointing the reason that the transaction on the batch is failing
 - Pointing all transactions that will make the batch fail

During this process we should review the following projects:
- Safe Ape - is a python project, so we won't be able to use it as is. We can get inspiration on some of the operations they allow to do. [Documents and repo available here](https://safe.ape.tax/)
- Tenderly - could we use Tenderly API to send the transaction parameters and show a transaction simulation for the user on Tenderly. Dima from mobile is already researching with Tenderly, we should sync with him on his discoveries
- Safe Simulator - this is a POC written by Richard where he uses Ganache to fork the chain and simulate in user browser. Code is available [in this repository](https://github.com/rmeissner/safe-simulator). There is also [an example](https://github.com/rmeissner/safe-simulator-app) of this library used on a Safe App environment

## Requirements
What we should expect as an outcome for this ticket

- [x] The three options above are reviewed.
- [x] A document is created comparing all three different approaches above, putting strengths and  weaknesses of each method
- [x] A final solution is proposed to be developed on the next roadmap cycle
- [x] A small POC is provided of any interesting functionality discovered during the investigation phase

## Links
Related problem statement https://github.com/gnosis/safe-pm/issues/41

mikheevm:
## What it solves
Resolves #200 

## How this PR fixes it
This PR integrates Tenderly Simulation API on the batch review screen. After the simulation, it shows if the transaction was successful and links to the simulation on Tenderly.

It provides a complete simulation with taking Safe guards into account by editing the threshold stored in the safe contract to 1 and providing a pre-validated signature from an owner. 

## How to test it
Create a new transaction batch, go to the review screen, and click simulate. For more detailed instructions, there's a demo below

## Screenshots
https://www.loom.com/share/44f37e9987ec4ac693d382681185a1dd


dasanra (approved):
I think it's ready for QA. Just in case as I changed some code would be nice to get another approve @DaniSomoza @yagopv

JagoFigueroa:
Hola señores, I think there is an edge case with the simulation when operating with non batched transactions (I might be wrong on the exact cause but I am able to reproduce it with non batched transactions only).

Basically if I create a single send tx via transaction builder (either a fallback method one from my safe to another safe or a transfer tx to a EOA wallet) of an asset amount that I don't have in my safe and should fail is marking it as all bueno:

Screenshot 2022-04-25 at 14 45 46


Screenshot 2022-04-25 at 14 32 13

but as soon as this invalid txs are batched the simulation is marking them as failed correctly:

Screenshot 2022-04-25 at 14 32 41

Un saludo

JagoFigueroa:
Opening a separate ticket for the issue mentioned above to make the PR review process a bit more convenient, will link the ticket here when I have it.

Un abrazo ;)

JagoFigueroa (approved):

JagoFigueroa:
All bueno from my side, bien trabajado compañero ;)

PR #405, opened 2022-03-28
fix(tx-builder): Unnecesary download batch modal

### LINKED ISSUE #397: [Tx builder] Batch download from edit batch screen should download it named
## Description
As it is possible to edit a batch name in this screen (and is required to have one before adding one) we should just download it when the download button is clicked instead of asking for a name for it. This would be the same behaviour as doing so From the transactions library screen.

## Steps to reproduce

1. 
Screenshot 2022-03-23 at 15 52 09

2. 
Screenshot 2022-03-23 at 15 52 30

## Expected result
Batch is downloaded

## Obtained result
Screenshot 2022-03-23 at 15 52 42

yagopv:
## What it solves
Resolves #397 

## How this PR fixes it
By removing the DownloadBatchModal and using the name of the batch directly


mikheevm (approved):

dasanra (approved):

JagoFigueroa:
All bueno ;)

Issue #398, opened 2022-03-23
[Tx builder] Limit the files that can be uploaded via drag and drop


JagoFigueroa:
## Description
It is possible to upload files different from .json when dragging and dropping and we should avoid this even if the file contents are valid. This behaviour is blocked when uploading a file from the explorer:
 
![Screenshot 2022-03-23 at 16 05 31](https://user-images.githubusercontent.com/92332710/159730709-d07dfcc0-1982-4785-ac5f-357a0cd87e1e.png)

but the ability to drag and drop makes it a bit trickier.

## Steps to reproduce
 1. Take a perfectly valid batch and change its extension to .doc
 2. Drag and drop it

## Expected result
Upload is blocked. 

Extra magic: 
Use this component
![Screenshot 2022-03-23 at 16 08 21](https://user-images.githubusercontent.com/92332710/159731336-cf8c215f-1b05-4b70-b228-1052d0cc8474.png)
to reflect unhappy path scenarios. Make it red if upload fails and prompt for a re-upload.

## Obtained result
![Screenshot 2022-03-23 at 16 07 32](https://user-images.githubusercontent.com/92332710/159731166-4ca84c0e-1bbf-4bf5-bdb6-6db7f1f2439a.png)

PR #406, opened 2022-03-28
fix(tx-builder): Feedback on drop non json files

yagopv:
## What it solves
Resolves https://github.com/gnosis/safe-react-apps/issues/398

## How this PR fixes it
onDrop non json files we show the user a message about the file should be JSON


DaniSomoza (approved):

JagoFigueroa:
Working like a charm! I say we change the message to: 'The uploaded file is not a valid JSON file' to give it a bit more swag

JagoFigueroa:
Muito goistoso. Obrigado!

Issue #400, opened 2022-03-23
[Tx builder] Firefox only, edge case - Line breaks on batch name edit when pressing the down arrow key


JagoFigueroa:
## Description

Very low priority issue as it is hard to reproduce. Selecting the batch name and copy / pasting to change it while clicking the down arrow key is causing some visual issues. The issue might be related to the text staying as 'selected' when the down arrow is pressed.

## Expected result
No line breaking.

## Obtained result
https://user-images.githubusercontent.com/92332710/159745490-9e3ce528-9319-48a0-ad73-22c1a8b04a08.mov

PR #407, opened 2022-03-28
fix(tx-builder): Content editable firefox bug

yagopv:
## What it solves
Resolves https://github.com/gnosis/safe-react-apps/issues/400

## How this PR fixes it
There is a bug in Firefox with contenteditable spans that makes copying the editable content copy the span itself as well

https://stackoverflow.com/questions/26057848/cut-paste-from-to-contenteditable-to-itself-does-not-work-in-firefox

We are changing the span to a div

### DIFF
--- a/apps/tx-builder/src/components/EditableLabel.tsx
+++ b/apps/tx-builder/src/components/EditableLabel.tsx
@@ -21,7 +21,7 @@ const EditableLabel = ({ children, onEdit }: EditableLabelProps) => {
 
 export default EditableLabel;
 
-const EditableComponent = styled.span`
+const EditableComponent = styled.div`
   display: block;
   white-space: nowrap;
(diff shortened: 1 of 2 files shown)


apps/tx-builder/src/components/EditableLabel.tsx
```diff
@@ -21,7 +21,7 @@ const EditableLabel = ({ children, onEdit }: EditableLabelProps) => {
 
 export default EditableLabel;
 
-const EditableComponent = styled.span`
+const EditableComponent = styled.div`
```
DaniSomoza, line 24: It seems that this change is causing this warning:
Captura de pantalla 2022-03-29 a las 11 14 10

JagoFigueroa:
All good, thanks!

Issue #399, opened 2022-03-23
[Tx builder] Delete button on edit batch screen working as 'discard unsaved changes'


JagoFigueroa:
## Description
The behaviour of the removal button present when editing a saved batch as a bit inconsistent. Seeing the button is easily assumed that we are going to remove this batch but the button is working as a 'discard unsaved changes' redirecting the user to the main screen.

## Steps to reproduce
 1. From transaction library screen, click a batch to edit it.
 2. Perform a couple of edits and click the remove button.

## Expected result

Most complete scenario I can think of is:

- Delete batch button working as a batch removal redirecting the user to the batch library screen when action is confirmed.

- The 'Back to Your Transaction Library' working as a back button with a confirmation pop up if there are unsaved changes that will be lost in the process of going back.

## Obtained result

![Screenshot 2022-03-23 at 16 57 52](https://user-images.githubusercontent.com/92332710/159742219-dd5cb061-e908-4b32-9e79-5f05cbe83594.png)

Clicking cancel changes are discarded and user is redirected to the main app screen.

PR #408, opened 2022-03-29
fix(tx-builder): Change popup texts for making the messages more clear

yagopv:
## What it solves
Resolves https://github.com/gnosis/safe-react-apps/issues/399

## How this PR fixes it
Changing the content of the delete popup messages because right now is a little confusing regarding if we are deleting the batch from the library or not


yagopv:
After talking with @DaniSomoza I'm only changing the messages for solving this issue. If we want to merge the `edit batch` view with the `create batch` one there are more implications and will imply a more deeper change @dasanra 

We can think about this change as a separate issue @dasanra

DaniSomoza (approved):

JagoFigueroa:
All bueno, gracias!

PR #409, opened 2022-03-30
feat: bump dependencies

yagopv:
## What it solves
Bump package versions and fix vulnerabilities

## How this PR fixes it
1) Updating dev dependencies mainly as others are up to date
2) Updating configuration in `config-overrides.js` files in order to fix breaking things because create-react-app 5 breaking changes See related link
3) Upgrading eslint and eslint action to support > 8

## Related links
- create-react-app breaking change. [node modules not supported](https://github.com/facebook/create-react-app/issues/11756). Need to add them manually or use browser libraries 
- [web3 and create-react-app](https://github.com/ChainSafe/web3.js#web3-and-create-react-app)
- console error ["Buffer is not defined"](https://stackoverflow.com/questions/68707553/uncaught-referenceerror-buffer-is-not-defined)
- [source map generation warnings](https://github.com/facebook/create-react-app/issues/11924) when upgrading to create-react-app 5


JagoFigueroa:
All bueno from my side ;)

Issue #390, opened 2022-03-21
[TX-Builder] Decoding of "tuple" input type is not working


dasanra:
## Overview
Currently we are showing input types as a helper to guide the user on how to fill input values in the transaction builder. It's working fine for the simple types but is failing for "tuple". No matter how you change the input value is not passing validation. We don't even decode which values should be used.

## Goals
Tuple type in solidity have a special treatment and usually comes with a `components` field decoding the expected values in the tuple. Order in `tuple` is also important so it's relevant that we try to show this `components` decoding instead of generic `tuple`, so users are able to fill the expected parameters. This should also fix the validation. Be aware that any of the tuple components could be a tuple also

## Requirements
- If we find that an input is type "tuple" try to show the `components` parameters, that would be types following same type definition structure. 
- **Don't forget** to show that this components should be wrapped into an array

## Links
`setAllocatorLimits` in this `mainnet` smart contract has type `tuple`
https://etherscan.io/address/0xb32ad041f23eafd682f57fce31d3ea4fd92d17af#code

Details for "tuple" in Solidity
https://docs.soliditylang.org/en/v0.7.6/abi-spec.html?highlight=tuple#handling-tuple-types

PR #415, opened 2022-03-31
Next release

dasanra:
## Added

- Update manifest standard to use `icon` property instead a custom one (#388)
- [TX-Builder] Transaction library, import and export (#348) (#349)
- 

## Fixed

- EOA Address not clearing ABI field (#359)
- Tuple type not working as expected (#390)

## Chore

-  Bump dependencies (#409)

### DIFF
--- a/apps/compound/public/manifest.json
+++ b/apps/compound/public/manifest.json
@@ -1,6 +1,7 @@
 {
   "name": "Compound",
   "description": "Money markets on the Ethereum blockchain",
+  "iconPath": "Compound.png",
   "icons": [
     {
       "src": "Compound.png",
--- a/apps/drain-safe/public/manifest.json
+++ b/apps/drain-safe/public/manifest.json
@@ -1,6 +1,7 @@
 {
   "name": "Drain Account",
   "description": "Transfer all your assets in batch",
+  "iconPath": "logo.svg",
   "icons": [
     {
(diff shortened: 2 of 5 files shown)


JagoFigueroa:
Regression plan for the release https://gno.testrail.io/index.php?/runs/view/1410&group_by=cases:section_id&group_order=asc

JagoFigueroa:
Regression is all green! 

Only comment that I have is that on tx history screen txs done via the PR branch are being displayed like a multisend call (the one above):
Screenshot 2022-04-04 at 09 16 52

Let me know if we would need to talk to the backend team or it is just something related to PR branches :)

JagoFigueroa (approved):

JagoFigueroa:
I can confirm that icons are visible from tx history screen. However, I think there is something a bit off with backend on Gnosis Chain. Look at how drains are initially displayed on Gnosis Chain:

1. I perform a drain & check tx history

Screenshot 2022-04-05 at 08 38 47
(Displayed as separate send txs)

2. I create a second drain & check tx history

Screenshot 2022-04-05 at 08 40 40
(second Drain displayed as separate send txs, first one pops up but the separate send txs for it persist)

3. I wait for a while and go back to tx history screen

Screenshot 2022-04-05 at 08 42 44
(All bueno)

PR #427, opened 2022-04-27
chore: configure prettier

dasanra:
## What it solves
We are not using Prettier to follow code style guidelines. Also the rules were not quite aligned with other repositories. We will try to align all repositories to use same code styling rules.

## How this PR fixes it
- Update prettier configuration file
- Add a task to use prettier before each commit on staged files
- Configure eslint to avoid rules issues
- Lint all current code


yagopv (approved):

mikheevm:
Why do we remove print width? Imo the default 80 is too small

mikheevm:
https://github.com/safe-global/safe-react/blob/dev/.prettierrc safe-react is still using 120

yagopv:
Not sure but I like the 80 option 😬. For me is more readable

dasanra:
> https://github.com/safe-global/safe-react/blob/dev/.prettierrc safe-react is still using 120

@mikheevm yes it was a push to something more readable, I wanted to experiment with the apps first before proposing this cutdown also in `safe-react`

For me as I usually work with lots of windows side to side it's much more convenient to avoid horizontal scroll. Also 80 is the old standard for programming :smile:

mikheevm:
What screens you all have?

yagopv:
haha, currently ultrawide indeed 😂

dasanra:
> What screens you all have? joy if you open a file with 500+ lines with 80 line rule, you get a minimal code vertically...

After discused internally we decided on testing with 100 chars length lines

PR #431, opened 2022-05-02
chore: new release

dasanra:
## Added
 - #200


JagoFigueroa:
Hola! Only one minor issue to comment from my side that I think we can approach as a separate ticket after the release. I see that on Volta and Aurora chains the simulation is not working (I am assuming there is no tenderly support for these chains and this is expected): 

Screenshot 2022-05-03 at 13 36 03

Maybe we could hide the button for these chains or show a generic error for these scenarios when there is no support for a specific chain after the button is clicked?

Cheers!

JagoFigueroa (approved):

PR #442, opened 2022-05-09
fix(tx-builder): Wrong abi input breaks tx-builder

### LINKED ISSUE #439: [Tx Builder] App breaking when an invalid ABI format is used
## Steps to reproduce
Go to Tx builder app and use values like: [{}] as a contract ABI.

## Expected result
App not crashing when an invalid ABI is used.

## Obtained result
https://user-images.githubusercontent.com/92332710/167138729-9136103a-f76a-475d-af3e-9f8e30a350f1.mov

yagopv:
## What it solves
Resolves #439 

## How this PR fixes it
We should check if the object representing ABI items have props


JagoFigueroa:
Todo perfecto, gracias!

PR #447, opened 2022-05-23
feat: Add test suite for check the Safe Apps liveness

yagopv:
## What it solves
Adding Cypress testing to the safe-apps-repo

## How this PR fixes it
1) Adding a test suite for checking the status for the safe apps (Test will be retried on fail twice for avoid flakiness)
2) Adding commands for working with iframes
3) Adding a github action for integration with github pipes


package.json
```diff
@@ -45,10 +47,13 @@
     "@typescript-eslint/eslint-plugin": "^5.21.0",
     "@typescript-eslint/parser": "^5.21.0",
     "assert": "^2.0.0",
+    "axios": "^0.27.2",
+    "dotenv": "^16.0.1",
```
dasanra, line 51: This dependency is not alphabetically ordered. I was trying to find because dotenv is already used in some packages, but I saw dotenv-cli some lines below

.github/workflows/safe-apps-check.yml
```diff
@@ -0,0 +1,44 @@
+name: Test Safe Apps
+on:
+  workflow_dispatch:
```
dasanra, line 3: I assume this event was for development purposes

dasanra (approved):
Everything looks good!

.github/workflows/safe-apps-check.yml
yagopv: yep!

package.json
yagopv: done!

PR #452, opened 2022-05-25
feat(tx-builder): e2e

yagopv:
## What it solves
Add e2e tests support for Safe Apps

### Transaction builder

- Test to create and send a batch
- Test to create and send a batch using an ENS name
- Test to create and send a batch starting from ABI only
- Test to create and send a batch using custom data
- Test not allowing to create a batch using a wrong address
- Test to upload a batch, save it to the library, download it and remove it
- Test a valid batch is simulated as "successful"
- Test that a wrong batch is simulated as "failed"

### Drain Safe

- Test to perform a drain
- Test to not allow a drain without assets selected
- Test to perform a drain of some assets


JagoFigueroa:

cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
yagopv, line 1: We can remove this if we are not using it

JagoFigueroa: Removed! gracias caballero

JagoFigueroa (approved):

PR #454, opened 2022-06-07
fix(tx-builder): Fix Tenderly simulations not working after reorder (drag&drop)

### LINKED ISSUE #453: [TX-Builder] Simulation data not updating after drag and drop
## Description
When getting to the review screen before sending a transaction batch we have the possibility to simulate. Simulation data is not updated when reordering transactions in batch so if we try simulation again we could get a wrong result, as we are sending previous data

## Environment
 - Browser:
 - Wallet: MetaMask
 - Safe:
 - Environment:
   - production (rinkeby)

## Steps to reproduce
 1. Go to TX-Builder and create a batch with more than 1 transaction
 2. Go to the review step
 3. Simulate the TX on Tenderly
 4. Reorder transactions in batch via drag and drop
 5. Try to simulate again and see that there wasn't a simulation change even if we order transactions on a way that can't be executed

## Expected result
Simulation also takes into account the new transaction order

## Obtained result
Simulation is not aware that the order of the transactions changed

yagopv:
## What it solves
Resolves #453 

## How this PR fixes it
We are using the useTransactions hook inside the useSimulation one for up to date information not depending on UI state changes

## How to test it
1) Add several transactions and simulate in the Review and confirm
2) After that reorder transactions and simulate again => New simulation should reflect the changes

### DIFF
--- a/apps/tx-builder/src/hooks/useSimulation.ts
+++ b/apps/tx-builder/src/hooks/useSimulation.ts
@@ -1,4 +1,3 @@
-import { BaseTransaction } from '@gnosis.pm/safe-apps-sdk'
 import { useCallback, useState, useMemo } from 'react'
 import { TenderlySimulation } from '../lib/simulation/types'
 import {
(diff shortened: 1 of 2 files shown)


JagoFigueroa:
All bueno from my side ;)

PR #455, opened 2022-06-08
fix(tx-builder): Fixed wrong value inherited when switch the contract method

### LINKED ISSUE #450: [Tx builder] value inherited when value name of a method is the same
## Description
When switching between different methods with the same value name, the value data (if filled) will be inherited by the second method.

## Steps to reproduce
 1. Go to Tx builder app on Rin and use this contract ABI: https://rinkeby.etherscan.io/address/0xc7017c2D369d518Df2bE11753573AA6005A83440#code
 2. Select this method and fill the address fieldScreenshot 2022-05-24 at 14 56 33
 3. Switch to this boolean methodScreenshot 2022-05-24 at 14 57 29

 ## Obtained result
Boolean inherits de input data that was used for the previous method but as it is different from true or false, UI is not able to handle that info.Screenshot 2022-05-24 at 14 49 40

## Expected result
We should think how to handle this data inheritance. 

Either stop it completely or adapt it for compatible data types only, ignoring the equal value names.

DaniSomoza:
## What it solves
Resolves #450

## How this PR fixes it

Now contract values are stored in its own object in the form state to avoid collisions between params with the same name and different types:
Captura de pantalla 2022-06-08 a las 18 20 59

### Added unit tests to `<SolidityForm />` component

I Added a unit test to make sure that this issue is fixed, [see this unit test](https://github.com/safe-global/safe-react-apps/pull/455/files#diff-34a23d5696630e55647d95ceb53fed8095cdab018dcca3f6b67a8dbbd6649bdaR100).

## How to test it

You can use the Basic test contract `0xD7b9B6c7DCbC7F95F06e1Bc1e7894530b066EB23` in `Rinkeby` because a lot of methods has the `newValue` param name for different types:

```
    function testBooleanValue(bool newValue) public {
        booleanValue = newValue;
    }

    function testAddressValue(address newValue) public {
        addressValue = newValue;
    }

```


JagoFigueroa:
All bueno from my side, thank you señores ;)

dasanra (approved):

mikhailxyz:
is the base branch correct? i feel like I've already reviewed that code

DaniSomoza:
> is the base branch correct? i feel like I've already reviewed that code

yep I updated the base branch to `development` and I fixed some conflicts. Now It seems to be ready for review.

Sorry!

apps/tx-builder/src/components/forms/fields/SelectContractField.tsx
```diff
@@ -18,6 +18,10 @@ const SelectContractField = ({
 }: SelectContractFieldTypes) => {
   return (
     <Select
+      id={name}
```
mikhailxyz, line 21: is it guaranteed that the name wil be unique across all elements of the page? if not I'd use testId instead

DaniSomoza: good catch!

now I am using the `id` prop from `Field` component instead of the `name`. because `name` prop
might not be unique across all elements of the page.

```
const SelectContractField = ({
  value,
  onChange,
  options,
  label,
  name,
  id,
}: SelectContractFieldTypes) => {
  return (
    <Select
      id={id}
      inputProps={{
        id: `${id}-input`,
      }}
      name={name}
```

PR #471, opened 2022-06-23
fix(tx-builder): uploaded batch files are saved as undefined

### LINKED ISSUE #468: [Tx-builder] Avoid downloading batches as undefined.json
## Description
Although a bit of an edge case, it is possible to download a batch as 'undefined.json' which is something we should avoid.

## Steps to reproduce
 1. Go to tx-builder app and choose a batch to upload
 2. After it has been uploaded, download it

## Expected result
Download it with the name it already has or if not possible align the behaviour with when a created non saved batch is downloaded and name the downloaded file 'Transactions Batch.json'

Example of this:

Screenshot 2022-06-21 at 15 14 05

## Obtained result
Batch is downloaded as 'undefined.json'

## Video with the process

https://user-images.githubusercontent.com/92332710/174808012-a01718df-8764-4edf-8d5a-4c7dea8132ff.mp4

mikhailxyz:
## What it solves
Resolves #468 
Also minor code fixes (use of `var`, unnecessary `async`, add .idea folder to gitignore)

## How this PR fixes it
The reason for the bug was that we pass the batchTitle as a component or a string and in the case of the component it used `props.children` as a name. But the passed component didn't have a `children` prop, only the `filename` prop. Since there are different types of components we pass as `batchTitle` I added if statements to check which one is exactly passed

## How to test it
Test uploading and downloading batches from screens where it is allowed (create a batch, edit batch, review, etc)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ build
 .env.development.local
 .env.test.local
 .env.production.local
-.eslintcache
 
 npm-debug.log*
 yarn-debug.log*
@@ -39,3 +38,6 @@ deployments/
 cypress/reports
 cypress/videos
 cypress/screenshots
+
+# intellij ide files
+.idea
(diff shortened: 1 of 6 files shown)


dasanra (approved):

JagoFigueroa:
All bueno, gracias señor ;)

PR #476, opened 2022-07-12
feat(tx-builder): detect if the provided contract is a proxy

### LINKED ISSUE #456: [EPIC][TX-Builder] Enable proxy detection
## Overview
After discussing that proxy contracts are not straight forward to use in the TX-Builder https://github.com/safe-global/safe-pm/issues/63 we decided to enable automatic check to detect when a contract is a proxy and offer the user the possibility of loading the implementation ABI 

## Goals
- Detect most common and standard proxy patterns
- Offer to change implementation ABI, do not change without noticing the user
- Allow user to review implementation address and check in block explorer
- Positive answer just replace the ABI, keep in other case.

## Requirements

- [ ] Proxy contract detect. There can be different patterns
- [ ] Detect if an ABI can be resolved in that implementation address
- [ ] Suggest the change to the user
- [ ] Swap when confirming

## Screens
imagen

 - Figma: https://www.figma.com/file/G7snLqBXWNPANcOj5MhiR7/Transaction-Builder?node-id=1704%3A9019

mikhailxyz:
## What it solves
Resolves #456 

## How this PR fixes it
It uses https://github.com/gnosis/evm-proxy-detection to check if the contract is a proxy and if it is one, it shows a modal asking the user which ABI to use

## How to test it
Testing is only possible in the dev environment at the moment:
https://safe-team.dev.gnosisdev.com/app/rin:0x10888911E6bD655d72C3081384C17EA7F205375A/apps?appUrl=https://pr476--safereactapps.review.gnosisdev.com/tx-builder/
You can play around with some proxy and non-proxy contracts

Also, there's one **caveat:**
It only detects safe proxies starting 1.1.0

## Demo
https://www.loom.com/share/936aa391fec14af1adb5f27568e142ed


apps/tx-builder/src/components/utils/Confirmable.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const [show, setShow] = React.useState(true)
+    const { resolve, ...rest } = props
+
+    const proceed = (value: boolean) => {
+      setShow(false)
+      resolve(value)
+    }
+
+    if (show) {
+      // @ts-ignore
```
mikhailxyz, line 29: would be great if someone could check if it's possible to get rid of `ts-ignore`, couldn't figure it out

apps/tx-builder/src/components/utils/Confirmable.tsx
```diff
@@ -0,0 +1,78 @@
...
+    mountingNode = document.body,
+  ) =>
+  (props: Omit<ComponentProps, keyof Confirmable>) => {
+    const wrapper = mountingNode.appendChild(document.createElement('div'))
+
+    const promise = new Promise<boolean>(resolve => {
+      try {
+        ReactDOM.render(
+          <ThemeProvider theme={theme}>
+            {/* @ts-ignore */}
```
mikhailxyz, line 49: would be great if someone could check if it's possible to get rid of `ts-ignore`, couldn't figure it out

apps/tx-builder/src/utils.ts
```diff
@@ -19,45 +19,6 @@ export const enum FETCH_STATUS {
   SUCCESS = 'SUCCESS',
   ERROR = 'ERROR',
 }
-
-export enum CHAINS {
```
mikhailxyz, line 23: it uses safe-apps-provider for RPC calls now

apps/tx-builder/src/pages/Dashboard.tsx
```diff
@@ -62,17 +36,76 @@ const Dashboard = (): ReactElement => {
...
-  const isAddressInputFieldValid = isValidAddress(address) || !address
+  const isAbiAddressInputFieldValid = !abiAddress || isValidAddress(abiAddress)
 
-  const contractHasMethods = contract && contract.methods.length > 0 && !isABILoading
+  const contractHasMethods =
+    abiStatus === FETCH_STATUS.SUCCESS && contract && contract.methods.length > 0
 
-  const isTransferTransaction = isValidAddress(address) && !abi && !isABILoading
-  const isContractInteractionTransaction = abi && contract && !isABILoading
+  const isTransferTransaction =
```
DaniSomoza, line 44: I think that there is an issue with the `isTransferTransaction` condition.

If I use a EOA address the form is not showed:

Captura de pantalla 2022-07-18 a las 14 35 28


Captura de pantalla 2022-07-18 a las 14 26 50

apps/tx-builder/src/pages/Dashboard.tsx
```diff
@@ -62,17 +36,76 @@ const Dashboard = (): ReactElement => {
...
   const showNewTransactionForm = isTransferTransaction || isContractInteractionTransaction
 
   const showNoPublicMethodsWarning = contract && contract.methods.length === 0
 
+  // The flow is the following:
+  // 1. We receive the address input, we check if it's a valid address
+  // 2. If it's a valid address, we check if it's a proxy
+  // 3. If it's a proxy, we get the implementation address and check if there's an ABI for it
+  // 4. If there's an ABI for the implementation address, we show the ABI dialog
+  const handleAddressInput = useCallback(
```
DaniSomoza, line 57: As a suggestion you can keep the same logic that we had before and just create an `useEffect`that only detects if its a proxy and show the popup like this:

```
  const {
    open: showProxyModal,
    openModal: openProxyModal,
    closeModal: closeProxyModalModal,
  } = useModal()

  useEffect(() => {
    if (isValidAddress(address) && !!abi) {
      // we check if a proxy is present
      const proxyAddress = detectProxyTarget()

      if (proxyAddress) {
        openProxyModal()
      }
    }
  }, [abi, address])
```

apps/tx-builder/src/pages/Dashboard.tsx
```diff
@@ -96,23 +129,23 @@ const Dashboard = (): ReactElement => {
...
 
-          {/* Address Input */}
+          {/* ABI Address Input */}
           <AddressInput
             id="address"
             name="address"
             label="Enter Address or ENS Name"
             hiddenLabel={false}
-            address={address}
+            address={abiAddress}
```
DaniSomoza, line 138: why `abiAddress` instead of `address` ?

apps/tx-builder/src/components/utils/Confirmable.tsx
```diff
@@ -0,0 +1,78 @@
...
+type ConfirmationProps = {
+  proceed: (value: boolean) => void
+  show: boolean
+}
+
+type Confirmable = {
+  resolve: (value: boolean) => void
+}
+
+const confirmable =
```
DaniSomoza, line 15: Do you think that this component is needed? we have modals like `DeleteBatchModal`, `SaveBatchModal` that works without this logic.

apps/tx-builder/src/utils.ts
DaniSomoza: are you sure that we can remove it?

the ens resolution is not working for me:

Captura de pantalla 2022-07-18 a las 14 52 04

apps/tx-builder/src/pages/Dashboard.tsx
mikhailxyz: because we can now have two separate addresses: one for the ABI and another one for implementation

apps/tx-builder/src/components/utils/Confirmable.tsx
mikhailxyz: Yes, they work but in much simpler flows. I wanted to have something similar to js' `confirm` built-in

apps/tx-builder/src/components/modals/ImplementationABIDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+} & ConfirmationProps
+
+const ImplementationABIDialog: React.FC<Props> = ({
+  networkPrefix,
+  implementationAddress,
+  blockExplorerLink,
+  proceed,
+}) => {
+  return (
+    <GenericModal
```
DaniSomoza, line 20: If you close the modal using the Cross icon, the input field gets stuck in an infinite loader state:

Captura de pantalla 2022-07-18 a las 15 00 58
Captura de pantalla 2022-07-18 a las 15 01 04

apps/tx-builder/src/pages/Dashboard.tsx
mikhailxyz: This is the first thing I tried to do, but it turned out the component into a complete mess of states and effects. One thing triggers one effect, that effect triggers another, and it becomes very hard to keep track of the current state. Also, I don't think the effect is the right way to go, let's compare two flows:
1. With an effect
```
1. We get an onChange event
2. We set the address in the state
3. We run an effect to potentially update the `address` state again
4. We update the `address` state again
```
2. Without an effect
```
1. We get an onChange event
2. We calculate new state based on the address in the event
3. We set new state
```

In my opinion, **2)** is more pragmatic and straightforward than having multiple effects triggering one another.

mikhailxyz: Also, I told you that the `AddressInput` component triggers duplicate events, so if we have chained but separate effects it adds even more complexity to prevent them from running unnecessarily

DaniSomoza: What is your opinion about to use a different state variable `implementationProxyAddress` to track the implementation address like I suggested in [this comment](https://github.com/safe-global/safe-react-apps/pull/476#discussion_r923356145):

the flow now could be:

```
With an effect
1. We get an onChange event
2. We set the address in the state
3. We run an effect to detect if a proxy is present
4. If true we update the `implementationProxyAddress` variable state
```

and then in the JSX we can do:

```
address={implementationProxyAddress || address}
```

apps/tx-builder/src/pages/Dashboard.tsx
mikhailxyz: Fixed senor

apps/tx-builder/src/pages/Dashboard.tsx
mikhailxyz: I don't think that by decoupling effects, we win something. Checking if an address is a proxy is closer to input validation than a side effect to me. Having a single effect run on `onChange` seems more logical. Having separate asynchronous effects which are separate but ultimately depend on one trigger (user input) create unnecessary complexity, makes it harder to work with, and leads to sneaky bugs (one example would be the `AddressInput` component)

apps/tx-builder/src/utils.ts
mikhailxyz: Turned out to be a bug with the `AddressInput`, for some reason it only does ens resolving when you change passed `address` value instead of reacting to users input 🤷 good catch senor

apps/tx-builder/src/components/modals/ImplementationABIDialog.tsx
mikhailxyz: Fixed, thanks, senor! Was handling the case with empty abi returned from the service incorrectly

DaniSomoza:
Now if you use a EOA address seems to be working fine but if I use my Safe: `rin:0x57CB13cbef735FbDD65f5f2866638c546464E45F` and click on `Keep Proxy ABI` the form is not rendered:
Captura de pantalla 2022-07-19 a las 11 45 50

DaniSomoza:
Thanks! 

I can see an small flicker visual [object, object] issue in the ABI:

Captura de pantalla 2022-07-19 a las 10 32 26

apps/tx-builder/src/pages/Dashboard.tsx
yagopv: I have some concerns about the amount of code in this change handler because in the future we probably will need to make more changes over as the address input is the starting point for the tx-builder flows.

I understand your concerns with effects but probably extracting this to a hook with the corresponding effects as @DaniSomoza suggested in the first comment will make it more maintainable .

DaniSomoza:
It seems that you are calling `interfaceRepo?.loadAbi()` multiple times, the first time in `handleAddressInput` and again if you clicks on `Use Implementation ABI` Button in the Modal

apps/tx-builder/src/pages/Dashboard.tsx
mikhailxyz: > I have some concerns about the amount of code in this change handler because in the future we probably will need to make more changes over as the address input is the starting point for the tx-builder flows.

I agree that this is a somewhat complex handler, but the logic is relatively straightforward and shouldn't be a problem to adjust. Not more complicated than trying to glue together four different effects to run a single algorithm.  

> I understand your concerns with effects but probably extracting this to a hook with the corresponding effects as @DaniSomoza suggested in the first comment will make it more maintainable .

I don't see it this way. In my eyes, it will not make it more maintainable but overcomplicate things and lead to unexpected bugs. I already mentioned an example of `AddressInput` where the same approach of splitting things into separate effects made the effects low-key dependent on one another and led to sneaky bugs.

Ultimately, what we want to do here, is to handle the user's change event. So, as a part of the handler, we run a few checks of the user input, and then we decide if we want to use this input or derive another value from the user's input. 

If we write a "pseudo-code", my solution is closer to this:
```
const [address, setAddress] = useState("")

const onChange = (e) => {
  const value = e.target.value

  const implementation = isProxy(value)
  if (implementation) {
    const implementationAbi = getImplementationAbi(implementation)
    if (implementationAbi) {
      setAddress(implementation)
    }
  } else {
    setAddress(value)
  }
}
```
The approach with effects looks like this:
```
const [address, setAddress] = useState("")
const [implementationAddress, setImplementationAddress] = useState("")

useEffect(() => {
  const implementation = isProxy(address)

  if (implementation) {
    setImplementationAddress(implementation)
  }
}, [address])

useEffect(() => {
  if (implementationAddress) {
    setAddress(implementationAddress)
  }

}, [implementationAddress])

const onChange = (e) => {
  setAddress(e.target.value)
}
```

I honestly do not see how the second approach is more maintainable or easier to understand. 
I can also quote some parts from the [react.js docs](https://beta.reactjs.org/learn/you-might-not-need-an-effect):

> You don’t need Effects to handle user events. For example, let’s say you want to send an /api/buy POST request and show a notification when the user buys a product. In the Buy button click event handler, you know exactly what happened. By the time an Effect runs, you don’t know what the user did (for example, which button was clicked). This is why you’ll usually handle user events in the corresponding event handlers.

> When you choose whether to put some logic into an event handler or an Effect, the main question you need to answer is what kind of logic it is from the user’s perspective. **If this logic is caused by a particular interaction, keep it in the event handler.** If it’s caused by the user seeing the component on the screen, keep it in the Effect.

Also, dan_abramov [tweet](https://twitter.com/dan_abramov/status/1281671006659260416) about `useEffect`:
> In hindsight, my personal conviction has become that if some effect can’t safely over-fire (e.g. fire twice instead of once), there is a problem. Usually things that can’t over-fire are related to user actions ("buy", "send", "finish"). Actions start in event handlers. Use them!

mikhailxyz:
> It seems that you are calling interfaceRepo?.loadAbi() multiple times with the same address, the first time in handleAddressInput and again in the useAbi() hook when you clicks on Use Implementation ABI Button in the Modal

Yes, this is true because we need to check if the ABI exists before triggering the modal, and there's no other way to check if it exists besides trying to fetch it. I don't think it's a big problem, it's not resource heavy and most likely gonna be cached after the first time

DaniSomoza:
It seems that if you set a proxy implementation address and then you type a EOA address the form state is not reset to normal EOA transfer:


https://user-images.githubusercontent.com/26763673/179795799-f60b8dc4-63d4-4af6-97a4-4791c831d584.mov

apps/tx-builder/src/components/modals/ImplementationABIDialog.tsx
```diff
@@ -0,0 +1,80 @@
...
+  position: relative;
+  padding: 24px;
+  max-width: 450px;
+`
+
+const StyledModalButtonsWrapper = styled(Box)`
+  margin-top: 24px;
+`
+
+const SEthHashInfo = styled(EthHashInfo)`
```
dasanra, line 74: Why not to use the "Styled" also here? I had trouble to read this. Thought it was a typo

apps/tx-builder/package.json
```diff
@@ -4,12 +4,15 @@
   "private": true,
   "homepage": "/tx-builder",
   "dependencies": {
-    "@gnosis.pm/safe-deployments": "^1.11.0",
-    "@gnosis.pm/safe-react-components": "^1.1.4",
+    "@gnosis.pm/safe-apps-provider": "^0.11.3",
+    "@gnosis.pm/safe-apps-sdk": "^7.5.0",
```
dasanra, line 8: This is already updated in the global repo package.json. Does it need to be added here?

apps/tx-builder/src/components/modals/ImplementationABIDialog.tsx
mikhailxyz: 🤷‍♂️ , fixed

apps/tx-builder/package.json
mikhailxyz: It depends on how we want to approach the SDK <> App relation, if we use a global dependency then any time we update the SDK globally, we would also have to update the apps. But, on the other hand, if it has a local dependency that doesn't match the global version number, it will install the required version specially for the app.

mikhailxyz: Hmm it looks like all the apps use the global version, will switch to the global here as well

mikhailxyz:
😅😅😅🫡🫡🫡

JagoFigueroa:
One pregunta señor @mikhailxyz 

1.1.1 safe addresses like `eth:0xB88F61E6FbdA83fbfffAbE364112137480398018` or `rin:0x55EaDd6984B73B1B78EaC57be81548e5d959D8a3` are detected as proxy contracts just fine. Is this magic expected or should only work for safes starting on 1.3.0 only?

Screenshot 2022-07-29 at 12 06 16
Screenshot 2022-07-29 at 12 06 35

mikhailxyz:
> One pregunta señor @mikhailxyz
> 
> 1.1.1 safe addresses like `eth:0xB88F61E6FbdA83fbfffAbE364112137480398018` or `rin:0x55EaDd6984B73B1B78EaC57be81548e5d959D8a3` are detected as proxy contracts just fine. Is this magic expected or should only work for safes starting on 1.3.0 only?
> 
> Screenshot 2022-07-29 at 12 06 16
> 
> Screenshot 2022-07-29 at 12 06 35

Oops, my mistake. I've just double-checked and it should work for safes >= 1.1.0, so the only version where it doesn't work is 1.0.0

JagoFigueroa:
Sin problema caballero ;) 

Bad news, looks like ABI resolution is not working as expected now:

https://user-images.githubusercontent.com/92332710/181768726-26eac7a2-2468-45f0-b68f-69eaa64a27d5.mp4

(left is the pr tx-builder and the right one the safe app from dev)

Leaving our fancy test contract in case you want to check with that one https://rinkeby.etherscan.io/address/0xD7b9B6c7DCbC7F95F06e1Bc1e7894530b066EB23#code but the ABI from a safe should do the trick as well.

mikhailxyz:
@JagoFigueroa lo arreglé, señor

JagoFigueroa:
Buenos días caballero, another pregunta for you. When using the implementation ABI for my safes I can see that the fallback method does not have the native token field making it not very usable. Is this expected? Should we do some magic with that?

https://user-images.githubusercontent.com/92332710/182090431-780998f1-b4c8-4690-b704-d066bc297f4a.mp4

mikhailxyz:
> https://safe-team.dev.gnosisdev.com/app/rin:0x10888911E6bD655d72C3081384C17EA7F205375A/apps?appUrl=https://pr476--safereactapps.review.gnosisdev.com/tx-builder/

The fallback method is not payable in the implementation ABI: https://github.com/safe-global/safe-deployments/blob/main/src/assets/v1.3.0/gnosis_safe.json#L316-L319

So, this is expected for now. Because it's not clear how to fix it, and if we want to fix it, I suggest fixing it in a different PR

mikhailxyz:
@JagoFigueroa, actually this is somewhat easy to fix by filtering out a non-payable fallback method, but since it'd touch general method filtering logic and require more testing, I would not do it here

JagoFigueroa:
Sin problema señor, we can always create a separate improvement for that :)

mikhailxyz:
@JagoFigueroa I've fixed both the nonpayable fallback and increased the width of the ABI modal

JagoFigueroa:
Enorme señor! Checking elegantemente ;)

JagoFigueroa:
Pop up is perfecto now but the nonpayable fallback is still no bueno unfortunately :(

For 1.1.0 safes is working nicely but for 1.3.0 and above the method is not visible anymore. We can tackle this perfectamente in a separate ticket to avoid any problems on this one.

mikhailxyz:
> for 1.3.0 and above the method is not visible anymore

Yes, that's expected. I've hidden it from the list of methods. The method is non-payable and there's nothing you can do with it.

JagoFigueroa:
Aaaaaaamigo. Then all bueno from my side, gracias señor!

PR #479, opened 2022-07-20
fix: e2e app liveness checks

JagoFigueroa:
## What it solves
Fix for the the new disclaimer for e2e liveness check for apps

### DIFF
--- a/cypress/integration/safe-apps-check.spec.js
+++ b/cypress/integration/safe-apps-check.spec.js
@@ -14,7 +14,7 @@ describe('Safe Apps List', () => {
       )
       const iframeSelector = `iframe[id="iframe-${safeApp.url}"]`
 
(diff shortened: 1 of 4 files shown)


cypress/integration/safe-apps-check.spec.js
```diff
@@ -14,8 +14,7 @@ describe('Safe Apps List', () => {
       )
       const iframeSelector = `iframe[id="iframe-${safeApp.url}"]`
 
-      cy.findByText('Accept all').click({ force: true })
```
yagopv, line 17: This is for the cookies not the modal

yagopv (changes requested):

DaniSomoza (approved):
Awesome! :3

PR #490, opened 2022-08-10
fix: e2e tests for proxy pop up

JagoFigueroa:
## What it solves
Adds step for proxy pop up for tests that use a safe address.


DaniSomoza (approved):
Approved! :3

PR #521, opened 2022-09-20
fix(safe-claiming-app): add block time desync buffer to amount calculation

schmanu:
## What it solves
The block time can lag behind the current (real) time and cause issues when estimating transactions because the actually vested amount is slightly smaller than computed in the front-end.

## How this PR fixes it
Substracts 20 seconds from the current timestamp when computing the vested amounts.

## How to test it
- on a safe which has most of the user airdrop claimed but still the ecosystem airdrop available
- create a transaction with a custom amount (i.e. 100) 
- Witness that the gas estimation does not fail anymore


katspaugh (approved):

apps/safe-claiming-app/src/utils/vesting.ts
```diff
@@ -4,9 +4,13 @@ import { VestingClaim } from "src/types/vestings"
...
 const EXPONENTIAL_CURVE = 1
 
+// Small buffer because the block time lags behind a little bit and this can cause failing transactions
+export const DESYNC_BUFFER = 20
+
 export const calculateVestedAmount = (vestingClaim: VestingClaim): string => {
   const durationInSeconds = vestingClaim.durationWeeks * 7 * 24 * 60 * 60
-  const timeStampInSeconds = Math.floor(new Date().getTime() / 1000)
+  const timeStampInSeconds =
+    Math.floor(new Date().getTime() / 1000) - DESYNC_BUFFER
```
usame-algan, line 13: As an alternative, could we also fetch the last confirmed block and use that timestamp instead of selecting the current time on clientside?

apps/safe-claiming-app/src/App.tsx
```diff
@@ -119,22 +120,24 @@ const App = (): ReactElement => {
 
   const chainConstants = CHAIN_CONSTANTS[safe.chainId]
 
-  const [delegates] = useDelegatesFile()
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [delegates, _, delegatesFileError] = useDelegatesFile()
```
usame-algan, line 124:
```suggestion
  const [delegates, , delegatesFileError] = useDelegatesFile()
```

This should also work

apps/safe-claiming-app/src/hooks/useAmounts.ts
```diff
@@ -9,23 +11,33 @@ export const useAmounts = (
   const [claimableAmount, setClaimableAmount] = useState("0")
   const [amountInVesting, setAmountInVesting] = useState("0")
   const [currentIntervalId, setCurrentIntervalId] = useState<number>()
+  const { safe, sdk } = useSafeAppsSDK()
+  const web3Provider = getWeb3Provider(safe, sdk)
```
usame-algan, line 15: Since this is not a hook I think we can move it into the useEffect and also add `safe` and `sdk` as dependencies.

PR #537, opened 2022-09-29
feat(wallet-connect): switch to the safe-apps-provider under the hood

### LINKED ISSUE #377: Adjust walletconnect app to support ethers `.wait` method
## Overview
If you submit a transaction with ethers.js and call the `wait` method to wait when the transaction is merged, the promise [will never resolve](https://ethereum.stackexchange.com/questions/123473/gnosis-safe-transaction-promises-fail-resolve). This can be worked around storing a faking transaction and using it with `eth_getTransactionByHash` as we do in `safe-apps-provider` https://github.com/gnosis/safe-apps-sdk/blob/master/packages/safe-apps-provider/src/provider.ts#L78-L90

## Goals
Support `ethers.js` wait in walletconnect Safe App

## Requirements
To be decided collectively. It can solved by migrating to safe-apps-provider or simply adjusting app code

## Links
more context: https://gnosisinc.slack.com/archives/CQB729LLF/p1647009035421679

mikhailxyz:
This came out as an attempt to fix #377

## What it solves
It switches the walletconnect Safe App to the `safe-apps-provider` package, so in case we change the handling of any of the RPC requests for the Safe Apps we can simply update it once in the SDK and bump the package version

## How to test it
Connect to your favourite app, transact with it and make sure everything is sound

## Why #377 wasn't solved
The problem is that in the case of walletconnect most of the wallet libraries are using the wallet provider only for WRITE requests, and READ requests are performed with a separate provider:
- Wagmi https://github.com/wagmi-dev/wagmi/blob/main/packages/core/src/connectors/walletConnect.ts#L133
- Onboard.js https://github.com/blocknative/web3-onboard/blob/v2-web3-onboard-develop/packages/walletconnect/src/index.ts#L244-L253
Therefore we cannot easily make it work with walletconnect. Further, I see three options:
1. Chase wallet libraries to adjust the behaviour and use the connector for read requests
2. Adjust .sendTransaction behaviour not to return safeTxHash right away and poll for the ethereum transaction hash
3. Do nothing and encourage apps to add Safe Apps support where this should work

I’m personally leaning towards 3) as the easiest option


apps/wallet-connect/src/hooks/useWalletConnect.tsx
```diff
@@ -56,93 +59,8 @@ const useWalletConnect = () => {
...
-                    value: txInfo.value || '0x0',
-                    data: txInfo.data || '0x',
-                  },
-                ],
-              })
-
-              result = safeTxHash
-              break
-            }
-            case 'gs_multi_send': {
```
dasanra, line 77: I supose this was a test method that wasn't still implemented?

apps/wallet-connect/src/hooks/useWalletConnect.tsx
```diff
@@ -56,93 +59,8 @@ const useWalletConnect = () => {
...
-            case 'eth_signTypedData':
-            case 'eth_signTypedData_v4': {
-              const [address, typedDataString] = payload.params
-              const typedData = JSON.parse(typedDataString)
-
-              if (!areStringsEqual(address, safe.safeAddress)) {
-                throw new Error('The address is invalid')
-              }
-
-              if (isObjectEIP712TypedData(typedData)) {
```
dasanra, line 131: We have this check here but we don't have it on the provider from the SDK. Could it be an issue?

mikhailxyz: The SDK method called by the provider has it: https://github.com/safe-global/safe-apps-sdk/blob/cef74f3ba70f1d5c98e51e83611408faf7982308/packages/safe-apps-sdk/src/txs/index.ts#L49-L51

So I'd say no

apps/wallet-connect/src/hooks/useWalletConnect.tsx
mikhailxyz: This is the method for batching transactions implemented by the OG app. I'm not sure if it's widely used. Quick GitHub search shows 1-2 apps: https://github.com/search?q=gs_multi_send&type=code

I think the mobile app continues to support it, so if we want to keep the support, we can implement it in the provider

apps/wallet-connect/src/hooks/useWalletConnect.tsx
```diff
@@ -56,93 +59,8 @@ const useWalletConnect = () => {
...
-              }
-            }
-            default: {
-              rejectWithMessage(wcConnector, payload.id, 'METHOD_NOT_SUPPORTED')
-              break
-            }
-          }
-
+          let result = await web3Provider.send(payload.method, payload.params)
+          console.log({ result })
```
dasanra, line 63: Remember to remove this 😄

mikhailxyz:
@dasanra there are some challenges related to the `gs_multi_send` migration to the provider, I outlined them in this ticket https://github.com/safe-global/safe-apps-sdk/issues/373

dasanra (approved):

PR #552, opened 2022-10-13
refactor(drain-safe): use Safe Apps provider on drain-safe

dasanra:
## What it solves
Uses Safe Apps provider on drain-safe


apps/drain-safe/src/hooks/useWeb3.ts
```diff
@@ -1,76 +1,23 @@
 import { useEffect, useState } from 'react'
 import Web3 from 'web3'
```
mikhailxyz, line 2: next step would be to refactor it to ethers

mikhailxyz (approved):

PR #553, opened 2022-10-19
fix(tx-builder): validate transactions in batch and add error message

mikhailxyz:
## What it solves
Resolves https://github.com/safe-global/safe-react-apps/issues/449

## How this PR fixes it
- The PR adds a validation step after loading the batch JSON. The validation checks that the transaction `value` field is not a number and that there are no numbers in `transaction.contractInputsValues`
- Moves `EditTransactionModal` into the `modals` folder
- Minor dependency updates

## How to test it
1. Export batch to a file
2. Change one of the fields to a number
3. Try to import the edited file. It shouldn't work and show an error instead

## Screenshots
image


apps/tx-builder/src/store/transactionLibraryContext.tsx
```diff
@@ -185,9 +212,10 @@ const TransactionLibraryProvider: React.FC = ({ children }) => {
     [initializeBatch],
   )
 
-  const importBatch = useCallback(
+  const importBatch: TransactionLibraryContextProps['importBatch'] = useCallback(
     async file => {
       const batchFile = await initializeBatch(await StorageManager.importBatch(file))
```
DaniSomoza, line 217: If you import an invalid batch file the error state is not removed, so if you update & re-import the batch file, nothing happens. I think that its fixed by removing the error state on import batch using `setErrorMessage(undefined)`:

```
  const importBatch: TransactionLibraryContextProps['importBatch'] = useCallback(
    async file => {
      setErrorMessage(undefined)
      const batchFile = await initializeBatch(await StorageManager.importBatch(file))

      return batchFile
    },
    [initializeBatch],
  )
```

mikhailxyz: I added it in `initializeBatch`: https://github.com/safe-global/safe-react-apps/pull/553/files#diff-12d2d5c28339b8ba03790e2e75a2644182e903793c6c71341913ece9b4059c1dR178

Can you reproduce this bug? I'll take a look if so

DaniSomoza: I think that its not related with this because we are already cleaning the error state in `initializeBatch` (line 178)...

maybe is something related with the `input` that we are using in the `CreateNewBatchCard` component, because Drag & Drop is working fine, but if you use the `input` directly it seems that it is not working...

DaniSomoza: ok it seems that we need to "reset" the value of the event target manually using `event.target.value = ''`:

```
  const handleFileSelected = (event: any) => {
    event.preventDefault()
    if (event.target.files.length) {
      onFileSelected(event.target.files[0])
      event.target.value = ''
    }
  }

```

see the first comment here: https://stackoverflow.com/questions/19643265/second-use-of-input-file-doesnt-trigger-onchange-anymore

mikhailxyz: Could you please record a video reproducing it? it is working fine for me:

https://www.loom.com/share/a847edb4ad5c4079bf9b305612eb2b23

DaniSomoza: > I added it in `initializeBatch`: https://github.com/safe-global/safe-react-apps/pull/553/files#diff-12d2d5c28339b8ba03790e2e75a2644182e903793c6c71341913ece9b4059c1dR178
> 
> Can you reproduce this bug? I'll take a look if so

You can reproduce it if you import an invalid file by clicking on "choose a file", after that you open the same file and fix it and then try to re-import it by clicking on "choose a file" again.

DaniSomoza: yep exactly, it seems that If you select the same file, the `onChange` event of `input[type=file]` will not fire.

DaniSomoza: https://user-images.githubusercontent.com/26763673/197175837-79afd439-6447-453d-bd55-3fd9636c2cc4.mov

mikhailxyz: I fixed it with your suggestion, señor. Good catch, thank you. Please check now if you still have this issue

DaniSomoza: Fixed! :3

Buen trabajo señor! 🚀

JagoFigueroa:
Testeado y verificado amigo, gracias!

PR #556, opened 2022-10-21
feat(e2e): Add more networks to e2e liveness tests

yagopv:
## What it solves
Adding more networks to the liveness tests

### DIFF
--- a/.github/workflows/safe-apps-check.yml
+++ b/.github/workflows/safe-apps-check.yml
@@ -1,5 +1,6 @@
 name: Test Safe Apps
 on:
+  push:
   workflow_dispatch:
     inputs:
       baseUrl:
--- a/.github/workflows/safe-apps-check.yml
+++ b/.github/workflows/safe-apps-check.yml
@@ -14,7 +14,7 @@ on:
       safeAddress:
         description: 'Safe address'
         required: true
-        default: '0xc322cde085A4C6EFc865E77eDDdF39c31262Fc70'
+        default: '0xfF501B324DC6d78dC9F983f140B9211c3EdB4dc7'
       configServiceBaseUrl:
         description: 'Config service base URL'
         required: true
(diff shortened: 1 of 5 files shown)


yagopv:
Hey @dasanra . Which networks should we add here. WDYT?

.github/workflows/safe-apps-check.yml
```diff
@@ -39,7 +42,143 @@ jobs:
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
           CYPRESS_BASE_URL: ${{ github.event.inputs.baseUrl || 'http://gnosis-safe.io/app' }}
           CYPRESS_NETWORK_PREFIX: ${{ github.event.inputs.networkPrefix || 'eth' }}
-          CYPRESS_TESTING_SAFE_ADDRESS: ${{ github.event.inputs.safeAddress || '0xc322cde085A4C6EFc865E77eDDdF39c31262Fc70' }}
+          CYPRESS_TESTING_SAFE_ADDRESS: ${{ github.event.inputs.safeAddress || '0xfF501B324DC6d78dC9F983f140B9211c3EdB4dc7' }}
+          CYPRESS_CONFIG_SERVICE_BASE_URL: ${{ github.event.inputs.configServiceBaseUrl || 'https://safe-client.gnosis.io' }}
```
mikhailxyz, line 46: This is not a config service URL but the client gateway senor. Also, should it use the new domain?

cypress/plugins/index.js
```diff
@@ -33,3 +33,15 @@ module.exports = async (on, config) => {
 
   return config
 }
+
+const NETWORK_ID = {
```
mikhailxyz, line 37: Could we maybe also pass the chain id?

.github/workflows/safe-apps-check.yml
yagopv: hehe yes, renamed

About the domain, web core prod is pointing to this yet 

Anyway when web-core goes live we need to redo several things here

JagoFigueroa (approved):

PR #559, opened 2022-10-25
feat(tx-builder): improve error message for overflow errors in tuples

### LINKED ISSUE #448: [TX-Builder] Tuple with long numbers cause overflow
## Description
When inputing a tuple of long numbers without wrapping as string they cause overflow.

This needs to be tested on an input that gets a tuple of u/int values.

## Environment
 - App: TX-Builder
 - Browser: All
 - Safe: All
 - Environment:
   - all

## Steps to reproduce
 1. Go to TX-Builder, using an ABI that requires uint, int or [uint], [int]
 2. Try to submit the form but an error will prevent with cause "overflow"

## Expected result
Transaction can be submited without issues

## Obtained result
Transaction is not possible to submit unless wrapping the values into strings

## Screenshots
imagen

mikhailxyz:
## What it solves
- Resolves #448. Similar to https://github.com/safe-global/safe-react-apps/pull/553, the root of the problem was `JSON.parse`
- Improves some types of errors in validation functions
- Minor dependency updates

## How this PR fixes it
It improves the error message and makes it straightforward for the user that numbers have to be encoded as strings. I saw this as the most straightforward solution compared to implementing a lossless custom JSON parser. The fix is very easy to do for the user.

## How to test it
1. Use `0x51A099ac1BF46D471110AA8974024Bfe518Fd6C4` contract on goerli
2. Select `getComplexTuple` method
3. Paste `["Hi", [1,2], 4, [6426191757410071111111, "hello", "bye", "0x6E4B36F8aE70BE64161E93C015a4201a37D6BEB3", true, 2]]` as a value

## Screenshots
Screenshot 2022-10-25 at 17 45 31


dasanra (approved):

JagoFigueroa:
Todo bueno, gracias!

PR #563, opened 2022-10-28
feat(wallet-connect): Wallet-connect v2 implementation

### LINKED ISSUE #178: [WalletConnect] Upgrade to WalletConnect v2
## Overview
WalletConnect v2 was released in the past days. It may worth to start creating a branch using the latest version of the dapp connector, as they expect to deprecate v1.
Estimated time for deprecation will be 6 months after v2 use surpass v, so no specific date for shutdown

## Goals
We add compatibility with WalletConnect v2 so we don't get chased by the v1 shutdown

## Requirements
 - Upgrade to WalletConnect v2
 - We have to test with other apps, specially if it's backwards compatible with Wallet Connect v2. We don't know if connection will work as expected with apps using an older version as connector.

## Links
In their docs they say that Wallets can detect to route v1 or v2 so it makes sense that we push into upgrade this faster in the Safe app
https://docs.walletconnect.com/migration#migration-schedule

DaniSomoza:
## What it solves
Resolves #178 

## How this PR fixes it

In the [WalletConnect v2.0 docs](https://docs.walletconnect.com/2.0/javascript/sign/wallet-usage#migrating-from-v1x) they  recommend to install `v1` and `v2` together for maximum compatibility. we have two different hooks  `useWalletConnectV1.tsx` & `useWalletConnectV2.tsx` to have both implementations decoupled and a `useWalletConnect.tsx` hook to use the correct version based on the [Walletconnect URI](https://docs.walletconnect.com/2.0/specs/core/pairing/pairing-uri).

the new WallectConnect v2.0 [needs a Project ID](https://docs.walletconnect.com/2.0/advanced/relay-server#project-id). I added a new `REACT_APP_WALLETCONNECT_PROJECT_ID ` env var.

Docs: 

- [Instalation Doc](https://docs.walletconnect.com/2.0/javascript/sign/installation)
- [WalletConnect v2.0 Protocol Specs](https://docs.walletconnect.com/2.0/specs/)

## How to test it

- DApp with Walletconnect version 1: [CowSwap](https://swap.cow.fi/#/swap?chain=goerli)
- DApp with Walletconnect version 2: [This test DApp](https://react-app.walletconnect.com/)

## Added Unit tests
Captura de pantalla 2022-12-14 a las 12 25 19
Captura de pantalla 2022-12-14 a las 12 26 31


apps/wallet-connect/src/hooks/useWalletConnectV2.tsx
```diff
@@ -0,0 +1,307 @@
...
+import {
+  NEW_SESSION_ACTION,
+  TRANSACTION_CONFIRMED_ACTION,
+  WalletConnectVersion,
+  WALLET_CONNECT_VERSION_2,
+} from '../utils/analytics'
+
+const { REACT_APP_WALLETCONNECT_PROJECT_ID, NODE_ENV } = process.env
+
+const isProduction = NODE_ENV === 'production'
```
mikhailxyz, line 18: nit: maybe it's worth exporting constants to a separate file like `constants.ts` to centralize access to environment variables

mikhailxyz: I'd also rename `REACT_APP_WALLETCONNECT_PROJECT_ID` to `REACT_APP_WALLETCONNECT_V2_PROJECT_ID` cos it's only relevant for v2

mikhailxyz (approved):

.github/workflows/deployment.yml
```diff
@@ -94,6 +95,7 @@ jobs:
           REACT_APP_TENDERLY_PROJECT_NAME: ${{ secrets.REACT_APP_TENDERLY_PROJECT_NAME }}
           REACT_APP_MMI_BACKEND_BASE_URL: ${{ secrets.MMI_BACKEND_BASE_URL }}
           REACT_APP_MMI_ENVIRONMENT: ${{ secrets.MMI_ENVIRONMENT }}
+          REACT_APP_WALLETCONNECT_PROJECT_ID: ${{ secrets.REACT_APP_WALLETCONNECT_PROJECT_ID_STAGING }}
```
yagopv, line 98: Is staging the env here?

apps/wallet-connect/src/App.tsx
```diff
@@ -62,19 +78,20 @@ const App = () => {
             <StyledCard>
               {connectionStatus === CONNECTION_STATUS.DISCONNECTED && (
                 <Disconnected>
-                  <WalletConnectField client={wcClientData} onConnect={data => wcConnect(data)} />
+                  <WalletConnectField wcClientData={wcClientData} wcConnect={wcConnect} />
```
yagopv, line 81: Why renaming the props here? the `onXXX` is the code convention in the other components

.github/workflows/deployment.yml
DaniSomoza: yep, this is the deploy env var for the dev & stage
```
  deploy:
    name: Deploy Dev/Staging
```

in the line 143 we are using the production one:
```
          react-app-walletconnect-project-id: ${{ secrets.REACT_APP_WALLETCONNECT_PROJECT_ID }}
```

apps/wallet-connect/src/hooks/useWalletConnectV2.tsx
DaniSomoza: done! I created a new constant `WALLETCONNECT_V2_PROJECT_ID` in the `constants.ts` file.

apps/wallet-connect/src/App.tsx
DaniSomoza: done!

yagopv (approved):

JagoFigueroa:
I think it would be neat to provide some feedback when the connection fails expectedly, like for example when trying to connect to a different chain.

https://user-images.githubusercontent.com/92332710/207603766-35fec01b-dd70-4fe1-810a-b4ace81c5c10.mp4

The connection fails as expected but maybe we could show a message too? The same for chains that are not compatible like solana, polkadot, etc. 

What do you think? :pepe-cheers:

DaniSomoza:
> 

Good Catch! right now we are only sending a message to the DApp, I will add a label to provide feedback to the user!

working on it...

DaniSomoza:
Done! buen señor! @JagoFigueroa 

Added feedback on invalid version 2 sessions! :3

Captura de pantalla 2022-12-16 a las 13 57 15

DaniSomoza:
@JagoFigueroa I can fix that in this PR!

DaniSomoza:
done! @JagoFigueroa

JagoFigueroa:
Espectacular trabajo buen señor, todo bueno 😉

PR #569, opened 2022-11-14
feat: Safe claiming widget

DiogoSoaress:
## What it solves
Resolves https://github.com/safe-global/web-core/issues/970

## How this PR fixes it
Creates a Widget to be displayed if `#widget` present in the app URL

## How to test it
Needs https://github.com/safe-global/web-core/pull/1149 to be run or merged

## Screenshots
_with claimed tokens_
Screenshot 2022-11-14 at 19 01 27
_without claimed tokens_
Screenshot 2022-11-15 at 12 36 42
_without voting power (not holding any)_
Screenshot 2022-11-15 at 20 31 18

### DIFF
--- a/apps/safe-claiming-app/src/Widget.tsx
+++ b/apps/safe-claiming-app/src/Widget.tsx
@@ -7,6 +7,7 @@ import {
   Typography,
   type TypographyProps,
   Link,
+  Skeleton,
 } from "@mui/material"
 import { BigNumber, ethers } from "ethers"
 import { useMemo } from "react"
@@ -60,10 +61,12 @@ const StyledButton = (props: ButtonProps) => (
   </Button>
 )
 
+const WIDGET_HEIGHT = 300
+
 const Widget = () => {
   const [delegates] = useDelegatesFile()
   const delegateAddressFromContract = useDelegate()
-  const balance = useTokenBalance()
+  const [balance, balanceLoading] = useTokenBalance()
   const { safe } = useSafeAppsSDK()
 
   const [vestings] = useAirdropFile()
@@ -84,17 +87,18 @@ const Widget = () => {
     }
   }, [delegateAddressFromContract, delegates])
 
-  const [userVestingStatus] = useFetchVestingStatus(
+  const [userVestingStatus, userVestingLoading] = useFetchVestingStatus(
     userVesting?.vestingId,
     userVesting?.contract
   )
 
-  const [ecosystemVestingStatus] = useFetchVestingStatus(
-    ecosystemVesting?.vestingId,
-    ecosystemVesting?.contract
-  )
+  const [ecosystemVestingStatus, ecosystemVestingLoading] =
+    useFetchVestingStatus(
+      ecosystemVesting?.vestingId,
+      ecosystemVesting?.contract
+    )
 
-  const [investorVestingStatus] = useFetchVestingStatus(
+  const [investorVestingStatus, investorVestingLoading] = useFetchVestingStatus(
     investorVesting?.vestingId,
     investorVesting?.contract
   )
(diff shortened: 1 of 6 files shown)


apps/safe-claiming-app/src/components/AppSwitch.tsx
```diff
@@ -0,0 +1,13 @@
+import App from "src/App"
+import Widget from "src/Widget"
+
+export const AppSwitch = () => {
+  const location = window.location
+  const hash = location.hash
+
+  if (hash === "#widget") {
```
iamacook, line 8: A hash is very generic. I would suggest a specific search parameter instead.

apps/safe-claiming-app/src/Widget.tsx
```diff
@@ -0,0 +1,191 @@
...
+    return BigNumber.from(0)
+  }, [balance, totalAllocation, totalClaimed])
+
+  const handleClickClaim = () => {
+    const url = `${window.location.ancestorOrigins[0]}/apps?safe=${
+      safe.chainId === 1 ? "eth" : "gor"
+    }:${safe.safeAddress}&appUrl=${CLAIMING_DATA_URL}`
+
+    // @ts-ignore
+    window.top.location.href = url
```
schmanu, line 116: I think it is quite dangerous, that this is currently possible from our Safe Apps as this is enables Safe apps to redirect our parent window. This could potentially lead to redirects to phishing / scam version of our interface from safe apps. 

In my opinion we should add some restrictions to our Safe apps by using the sandbox attribute of the iframe.
Some more info: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox
e.g.
```html
<iframe sandbox="allow-scripts" />
```

apps/safe-claiming-app/src/Widget.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  const ctaWidget = (
+    <SpaceContent>
+      <div>
+        <Title>Become the part of Safe future!</Title>
+        <Subtitle>
+          Holding $SAFE is required to participate in governance.
+        </Subtitle>
+      </div>
+      <StyledButton>Buy Tokens</StyledButton>
```
schmanu, line 127: I would remove this button for now. I don't know if we want to advertise buying the token. 
Plus it is currently not transferable.

apps/safe-claiming-app/src/hooks/useTokenBalance.ts
```diff
@@ -0,0 +1,41 @@
...
+import { BigNumber } from "ethers"
+import { useEffect, useState } from "react"
+import { CHAIN_CONSTANTS } from "src/config/constants"
+import { SafeToken__factory } from "src/types/contracts/factories/SafeToken__factory"
+import { getWeb3Provider } from "src/utils/getWeb3Provider"
+
+/**
+ * Fetches the current token balance.
+ */
+export const useTokenBalance = () => {
```
schmanu, line 11: Let's add a small test for this

apps/safe-claiming-app/src/hooks/useTokenBalance.ts
```diff
@@ -0,0 +1,41 @@
...
+        console.error(error)
+      }
+    }
+    if (chainConstants) {
+      fetchTokenBalance()
+    }
+    return () => {
+      isMounted = false
+    }
+  }, [chainConstants, safe.safeAddress, web3Provider])
```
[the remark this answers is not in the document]
DiogoSoaress: Good catch! Thanks!

apps/safe-claiming-app/src/hooks/useTokenBalance.ts
DiogoSoaress: Added a test

apps/safe-claiming-app/src/Widget.tsx
DiogoSoaress: I disabled the button and change the secondary text (see updated PR description)

apps/safe-claiming-app/src/Widget.tsx
```diff
@@ -0,0 +1,196 @@
...
+  return (
+    <Box
+      display="flex"
+      flexDirection="column"
+      alignItems="center"
+      height="300px"
+      p={3}
+      sx={{ backgroundColor: "background.paper" }}
+    >
+      {votingPower.eq(0) ? ctaWidget : votingPowerWidget}
```
schmanu, line 191: While loading data, the app flickers and briefly shows "Become part of the Safe future".

I would suggest that we use the loading states of the used hooks and probably introduce a loading state into the useTokenBalance hook to display Skeletons while loading.

schmanu (approved):
The flickering is gone and it worked when I tested the widget.

If we are aligned on including the widget in the app, we can merge IMO

DiogoSoaress:
Closing in favor of #575 once it branched from this branch and contains some extra fixes (hook refactor, votingPower fixed)

PR #572, opened 2022-11-15
feat(tx-builder): Add Transaction builder analytics

yagopv:
## What it solves
Resolves https://github.com/safe-global/web-core/issues/1126

## How this PR fixes it
We are adding some analytics to the Transaction Builder

## How to test it
Check the following Analytics are being sent using the correct [web-core version](https://github.com/safe-global/web-core/pull/1066)

1) 'Submit transactions confirmed' after a transaction is confirmed
2) 'Saved batch' after saving a batch
3) 'Remove batch' after removing a batch
4) 'Update batch' after updating a batch
5) 'Download batch' after pressing the download batch button
6) 'Import batch' after dragging or opening a file containing a batch

### DIFF
--- a/apps/tx-builder/src/lib/analytics.ts
+++ b/apps/tx-builder/src/lib/analytics.ts
@@ -1,0 +1,13 @@
+const SAFE_APPS_ANALYTIC_CATEGORY = 'safe-apps-analytics'
+
+export const trackSafeAppEvent = (action: string, label?: string) => {
+  window.parent.postMessage(
+    {
+      category: SAFE_APPS_ANALYTIC_CATEGORY,
+      action,
+      label,
+      safeAppName: 'Transaction Builder',
+    },
+    '*',
+  )
+}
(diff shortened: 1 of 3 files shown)


yagopv:
Adding some analytics to Transaction Builder in order to resolve https://github.com/safe-global/web-core/issues/1126. You can take a look to the description to see what's added.

Please tell me what more i can add

@clemoon @dasanra

dasanra (approved):

apps/tx-builder/src/lib/analytics.ts
```diff
@@ -0,0 +1,13 @@
...
+  window.parent.postMessage(
+    {
+      category: SAFE_APPS_ANALYTICS_CATEGORY,
+      action,
+      label,
+      safeAppName: 'Transaction Builder',
+    },
+    '*',
+  )
+}
```
mikhailxyz, line 13: IMO this function needs to be exported to a library and then imported into apps so the app doesn't have to define how and what it sends

yagopv: yep agree, but let's see if we are going to use it more because right now we have it in the Wallet Connect app but other apps went out of scope right now from the current analytics requirements. I merged this because it was already done and could be useful

PR #580, opened 2022-11-22
feat(safe-claiming-app): Dark theme

DiogoSoaress:
## What it solves
Dark mode for Claiming app + widgets

Depends on https://github.com/safe-global/web-core/pull/1219 to be tested

## How this PR fixes it
Reads a dark theme flag from the appUrl with a fallback to the system preference

## How to test it
Two ways to test the dark mode:
  - Set your system preferences to dark
  - Needs to run https://github.com/safe-global/web-core/pull/1219 and this branch locally.
Observe the Governance widgets load according to the web-core app theme.
The Claiming app has dark mode applied.

## Screenshots
Screenshot 2022-11-22 at 13 19 26


apps/safe-claiming-app/src/hooks/useDarkMode.ts
```diff
@@ -0,0 +1,32 @@
...
+  return window.matchMedia("(prefers-color-scheme: dark)").matches
+}
+
+export const useDarkMode = (): boolean => {
+  const location = window.location
+
+  const [isDarkMode, setIsDarkMode] = useState<boolean>(false)
+
+  useEffect(() => {
+    setIsDarkMode(location.hash.endsWith("+dark") ?? isSystemDarkMode())
```
schmanu, line 15: For me this does not work because it always sets it to `false` if there is no +dark in the hash, even with system dark mode true. Either use `||`instead of `??` or this should rather be done via query parameter.
Alternatively we could only support SystemDarkMode until darkmode settings are passed through the safe-apps-sdk

DiogoSoaress: are you running https://github.com/safe-global/web-core/pull/1219 locally? I included the way to test it in the Description

mikhailxyz:
Please make sure that PR names follow conventional commits [guidelines](https://www.conventionalcommits.org/en/v1.0.0/) and that the scope of the change corresponds to the app's name defined in package.json

This is needed for our automated versioning & changelog tooling to work.

apps/safe-claiming-app/src/components/steps/Claim/SelectedDelegate.tsx
```diff
@@ -16,20 +17,37 @@ export const SelectedDelegate = ({
...
+                    : isDarkMode
+                    ? `${palette.background.paper} !important`
+                    : `${palette.border.background} !important`,
+                border: isDarkMode
+                  ? ({ palette }) =>
+                      isHover
+                        ? `1px solid ${palette.safeGreen.main}`
+                        : `1px solid ${palette.text.primary}`
+                  : undefined,
+              }}
```
usame-algan, line 50: This is a lot of logic for a small design difference. I would suggest either removing it or moving it into the theme overrides.

apps/safe-claiming-app/src/components/steps/Claim/index.tsx
```diff
@@ -173,7 +176,7 @@ const Claim = ({ handleBack, state, handleUpdateState, handleNext }: Props) => {
         />
         <Typography variant="subtitle2">
           Awarded total allocation is{" "}
-          <span style={{ color: "#121312" }}>
+          <span style={{ color: palette.text.primary }}>
```
usame-algan, line 179: Could we do `<Typography color="text.primary">` instead?

apps/safe-claiming-app/src/components/steps/Distribution/index.tsx
```diff
@@ -51,6 +51,7 @@ const Distribution = ({ handleNext, handleBack }: Props) => {
             <PercentageWrapper
               sx={{
                 backgroundColor: ({ palette }) => palette.safeGreen.main,
+                color: "#121312",
```
usame-algan, line 54: For static values, you could also use the color palette directly i.e.

```
import palette from "./colors"
...
color: palette.text.primary
```

apps/safe-claiming-app/src/components/steps/Distribution/index.tsx
```diff
@@ -70,6 +71,7 @@ const Distribution = ({ handleNext, handleBack }: Props) => {
             <PercentageWrapper
               sx={{
                 backgroundColor: ({ palette }) => palette.secondary.light,
+                color: "#121312",
```
usame-algan, line 74: See above comment

apps/safe-claiming-app/src/components/steps/Distribution/index.tsx
```diff
@@ -85,6 +87,7 @@ const Distribution = ({ handleNext, handleBack }: Props) => {
             <PercentageWrapper
               sx={{
                 backgroundColor: ({ palette }) => palette.info.main,
+                color: "#121312",
```
usame-algan, line 90: See above comment

apps/safe-claiming-app/src/components/steps/SafeInfo/index.tsx
```diff
@@ -1,16 +1,28 @@
...
 
 type Props = {
   handleBack: () => void
   handleNext: () => void
 }
 
+const InfoCard = styled(Card)<{ darkMode: boolean }>(({ darkMode, theme }) => ({
+  backgroundColor: darkMode
+    ? theme.palette.background.paper
+    : theme.palette.safeGreen.light,
```
usame-algan, line 17: I would suggest using `safeGreen.light` for both dark and light mode.

apps/safe-claiming-app/src/config/theme.ts
```diff
@@ -1,173 +1,182 @@
...
+      },
+      MuiButton: {
+        styleOverrides: {
+          sizeLarge: {
+            padding: "10px 40px",
+          },
+          contained: {
+            border: "1px solid transparent",
+            "&:hover": {
+              backgroundColor: "transparent",
```
usame-algan, line 142: This seems different to the buttons in web-core where it keeps the background color on hover and just makes it a bit darker. I would suggest keeping it consistent.

usame-algan (changes requested):
The button on the success page looks different

Screenshot 2022-11-24 at 11 52 52

apps/safe-claiming-app/src/components/steps/Claim/SelectedDelegate.tsx
schmanu: I'll try to somehow combine it in a way that it looks decent in dark and light mode but works the same otherwise

apps/safe-claiming-app/src/components/steps/SafeInfo/index.tsx
schmanu: I put `safeGreen.light` for both in a first version, but lila explicitly wanted it to be colored like this. 
She marked everything which was important for her with these red boxes.

https://www.figma.com/file/igarLtnMhVtrYNOATmjJDc/Claim-S-Tokens-App?node-id=2363%3A15272&t=OB4Zd1XArFbKL8p3-0

apps/safe-claiming-app/src/components/steps/Distribution/index.tsx
DiogoSoaress: we can call just
```css
color: 'text.primary'
```
without the import, no?

schmanu: No because those Percentage colors are the same in light and dark mode, but text.primary is not the same in both modes.

usame-algan (approved):
Looks really cool 👍

PR #588, opened 2022-11-25
feat(safe-claiming-app): Parse theme from URL hash

DiogoSoaress:
## What it solves
Part of https://github.com/safe-global/web-core/issues/1221

## How this PR fixes it
The `useDarkMode` hook reads the theme from the URL's hash instead of from the URL search param

### DIFF
--- a/apps/safe-claiming-app/src/hooks/useDarkMode.ts
+++ b/apps/safe-claiming-app/src/hooks/useDarkMode.ts
@@ -12,7 +12,13 @@ export const useDarkMode = (): boolean => {
   const [isDarkMode, setIsDarkMode] = useState<boolean>(false)
 
   useEffect(() => {
-    setIsDarkMode(location.hash.endsWith("+dark") || isSystemDarkMode())
+    const isDark = location.hash.endsWith("+dark")
+      ? true
+      : location.hash.endsWith("+light")
+      ? false
+      : isSystemDarkMode()
+
+    setIsDarkMode(isDark)
   }, [location.hash])
 
   return isDarkMode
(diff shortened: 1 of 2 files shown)


apps/safe-claiming-app/src/hooks/useDarkMode.ts
```diff
@@ -12,11 +12,8 @@ export const useDarkMode = (): boolean => {
   const [isDarkMode, setIsDarkMode] = useState<boolean>(false)
 
   useEffect(() => {
-    const urlParams = new URLSearchParams(location.search)
-    const urlDarkMode = urlParams.get("theme")
-
-    setIsDarkMode(urlDarkMode === "dark" || isSystemDarkMode())
-  }, [location.search])
+    setIsDarkMode(location.hash.endsWith("+dark") || isSystemDarkMode())
```
katspaugh, line 15: We shouldn't use `isSystemDarkMode`. First of all, web-core will always send +dark if needed, and second of all, the web-core setting can be different than the system setting.

katspaugh (changes requested):

apps/safe-claiming-app/src/hooks/useDarkMode.ts
katspaugh: If you want to use the system fallback for the main claiming app, then the widget should explicitly send +light. Otherwise, you'll always fallback to the OS setting which might not be correct.

DiogoSoaress: I think we don't need to fallback to the system setting at all. I'm gonna remove it given web-core always explicitly passes dark / light for the theme.

DiogoSoaress: I now explicitly check if `+light` is passed. If none of the theme preferences is passed in the URL then we are rendering the main Claiming app which uses the system preference

katspaugh (approved):

PR #604, opened 2022-12-02
fix(wallet-connect): fixed open safe app redirection

DaniSomoza:
## What it solves
Resolves  "Open Safe App" redirection in the new Web-core ui


yagopv (approved):

JagoFigueroa:
Hola! Small issue:


https://user-images.githubusercontent.com/92332710/205263505-81737204-2566-4809-838a-5e6d680fe9b7.mp4


I think we should avoid the "infinite loader" on the original tab :peepo-shy:

DaniSomoza:
Thanks @JagoFigueroa I removed the loading state when we perform a redirection because its no needed now! :3

mikhailxyz (approved):

JagoFigueroa:
Todo bueno, gracias! 😉

PR #612, opened 2022-12-12
fix(safe-claiming-app): correct title color

schmanu:
## What it solves
Correct title color of claiming widget


DiogoSoaress (approved):

schmanu:
@liliya-soroka: Merging this without QA as it only changes the text color and nothing else.

PR #614, opened 2022-12-12
feat(e2e): Update e2e tests for work with web-core

yagopv:
## What it solves

Update e2e tests for work with web-core


index.html
```diff
@@ -0,0 +1,4741 @@
+<html lang="en">
```
dasanra, line 1: Is this extra big file needed in the repo? It seems something autogenerated

yagopv: noooope 😱 removed, thanks

dasanra (approved):

PR #615, opened 2022-12-12
fix(tx-builder): Added clipboard permission to the `manifest.json` file

### LINKED ISSUE #568: [Tx-builder] Add clipboard permissions
## Description
We should add clipboard write permissions to the tx-builder safe app the same way we have done for 'SIWE' so that the copy to clipboard' feature can be only used when the user has agreed to do so.

This button can be found (for example) in the review batch screen:

Screenshot 2022-11-11 at 15 55 24

DaniSomoza:
## What it solves
Resolves #568 

## How this PR fixes it

Added `clipboard-write` permission in the `manifest.json` file

```
{
  "name": "Transaction Builder",
  "description": "A Safe app to compose custom transactions",
  "iconPath": "tx-builder.png",
  "icons": [
    {
      "src": "tx-builder.png",
      "sizes": "256x256",
      "type": "image/png"
    }
  ],
  "safe_apps_permissions": ["clipboard-write"]
}

```

## How to test it

Go to the current RC And check that you can copy the addresses into your clipboard

RC: https://release120--webcore.review-web-core.5afe.dev/

### DIFF
--- a/apps/tx-builder/public/manifest.json
+++ b/apps/tx-builder/public/manifest.json
@@ -8,5 +8,6 @@
       "sizes": "256x256",
       "type": "image/png"
     }
-  ]
+  ],
+  "safe_apps_permissions": ["clipboard-write"]
 }


mikhailxyz (approved):

apps/tx-builder/public/manifest.json
```diff
@@ -8,5 +8,6 @@
       "sizes": "256x256",
       "type": "image/png"
     }
-  ]
+  ],
+  "safe_apps_permissions": ["clipboard-write"]
```
mikhailxyz, line 12: Shouldn't it be camelCase? https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/permissions

yagopv: We are using here a custom property different than the standard one as we want the different directives allowed in the Feature-Policy API and we just translate them to the allow attribute in the iframe.

The permissions standard one i think allow another subset and have implications in the browser behavior

JagoFigueroa:
Checking with señor @DaniSomoza we see that the clipboard permission preferences are ignored and copying is always allowed regardless they are set to true or false. There might be some more generic iframe permissions in place and always set to allow overwriting this one. 

We might need to do some more research for this.

yagopv (approved):

JagoFigueroa:
This ticket can be closed. Due to the reasons mentioned in the comment above we won't be implementing this yet.

PR #623, opened 2022-12-15
chore: release with fix for loading state in safe-claiming-app

schmanu:
## What it solves
Fixes error in current main version where the app showed the "No Airdrop" screen for unredeemed allocations before the deadline was reached.

## How this PR fixes it
Fixes a bug in the loading state.

## How to test it
Use a Safe with unredeemed tokens on mainnet: It now correclty displays the claiming steps.

### DIFF
--- a/apps/safe-claiming-app/src/App.tsx
+++ b/apps/safe-claiming-app/src/App.tsx
@@ -98,11 +98,11 @@ const App = (): ReactElement => {
 
   const [safeTokenAllocation, allocationError, isVestingLoading] =
     useSafeTokenAllocation()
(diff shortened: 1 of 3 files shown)


yagopv (approved):
Could you tests this @JagoFigueroa

JagoFigueroa:
Checked on mainnet, all good 😄

JagoFigueroa:
Seeing that the live version was rolled back correctly I think it is better we do a small regression for it first, we will notify here when it is finished 😄

liliya-soroka:
Ready to go live

PR #639, opened 2023-01-03
style(safe-claiming-app): claiming app minor styles

DiogoSoaress:
## What it solves
Resolves https://github.com/safe-global/safe-react-apps/issues/627

## How this PR fixes it
1. Gives the "Claim and delegate" button same height as the input field
2. Make "Voting Power" and "Delegate" parts of claiming widget clickable and navigate to the claiming app

## How to test it
1. Go to Claiming widget in Dashboard.
2. Click Safe icon or tokens value and you should navigate to the Claiming app
3. Click on the Delegate and you should navigate to the Claiming app

## Screenshots
Screenshot 2023-01-04 at 10 35 52

### DIFF
--- a/apps/safe-claiming-app/src/components/steps/Claim/index.tsx
+++ b/apps/safe-claiming-app/src/components/steps/Claim/index.tsx
@@ -236,7 +236,7 @@ const Claim = ({ handleBack, state, handleUpdateState, handleNext }: Props) => {
                   variant="contained"
                   size="large"
                   disableElevation
-                  sx={{ width: 1, paddingX: 0 }}
+                  sx={{ width: 1, paddingX: 0, height: "47px" }}
                   disabled={buttonDisabled}
                   onClick={claimTokens}
                 >
--- a/apps/safe-claiming-app/src/widgets/ClaimingWidget.tsx
+++ b/apps/safe-claiming-app/src/widgets/ClaimingWidget.tsx
@@ -89,6 +89,14 @@ const StyledButton = (props: ButtonProps) => (
   </Button>
 )
 
+const StyledButtonLink = styled(Button)`
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: 8px;
+  margin-bottom: 8px;
+`
+
 const WIDGET_HEIGHT = 300
 
 const ClaimingWidget = () => {
@@ -140,29 +148,19 @@ const ClaimingWidget = () => {
     <>
       <div>
         <Title>Your voting power</Title>
-        <Link
+        <StyledButtonLink
           href={claimingSafeAppUrl}
-          rel="noopener noreferrer"
-          target="_blank"
-          underline="none"
+          sx={{ "&:hover": { backgroundColor: "secondary.light" } }}
         >
-          <Box
-            display="flex"
-            alignItems="center"
-            justifyContent="center"
-            gap={1}
-            mb={1}
-          >
-            <SafeIcon />
-            <Typography variant="h5" color="text.primary">
-              {votingPower ? (
-                formatAmount(Number(ethers.utils.formatEther(votingPower)), 2)
-              ) : (
-                <Skeleton />
-              )}{" "}
-            </Typography>
-          </Box>
-        </Link>
+          <SafeIcon />
+          <Typography variant="h5" color="text.primary">
+            {votingPower ? (
+              formatAmount(Number(ethers.utils.formatEther(votingPower)), 2)
+            ) : (
+              <Skeleton />
+            )}{" "}
+          </Typography>
+        </StyledButtonLink>
       </div>
       {totalClaimed?.gt(0) ? (
         <>
@@ -186,7 +184,10 @@ const ClaimingWidget = () => {
                 target="_blank"
                 underline="none"
               >
-                <SelectedDelegate delegate={currentDelegate} />
+                <SelectedDelegate
+                  delegate={currentDelegate}
+                  onClick={() => {}}
+                />
               </Link>
             </Box>
           )}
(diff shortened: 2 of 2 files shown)


apps/safe-claiming-app/src/components/steps/Claim/index.tsx
```diff
@@ -236,7 +236,7 @@ const Claim = ({ handleBack, state, handleUpdateState, handleNext }: Props) => {
                   variant="contained"
                   size="large"
                   disableElevation
-                  sx={{ width: 1, paddingX: 0 }}
+                  sx={{ width: 1, paddingX: 0, height: "47px" }}
```
schmanu, line 239: I don't know whats with the buttons in the claiming app.
I would change it by doing the following:

1. remove `size="large"`
2. fix the styles in the theme.ts (currently the buttons are way too small):
This seems to work:
```ts
      MuiButton: {
        styleOverrides: {
          sizeLarge: {
            fontSize: "16px",
            padding: "12px 40px",
          },
          sizeMedium: {
            fontSize: "16px",
            padding: "12px 24px",
          },
          root: ({ theme }) => ({
            borderRadius: theme.shape.borderRadius,
            fontWeight: "bold",
            lineHeight: 1.25,
            borderColor: theme.palette.primary.main,
            textTransform: "none",
            "&:hover": {
              boxShadow: "none",
            },
          }),
        },
      },
```

3. customize the padding to slightly smaller so that the button text does not wrap (8px for instance works for me)

apps/safe-claiming-app/src/widgets/ClaimingWidget.tsx
```diff
@@ -140,22 +140,29 @@ const ClaimingWidget = () => {
     <>
       <div>
         <Title>Your voting power</Title>
-        <Box
-          display="flex"
-          alignItems="center"
-          justifyContent="center"
-          gap={1}
-          mb={1}
+        <Link
```
schmanu, line 143: To me it's a bit surprising that the Voting power is now clickable and leads to the claiming app.

At least without any visual indicator that this is clickable. Probably we should use `ButtonBase` or something similar as a wrapper instead of link to clarify that this is clickable?

apps/safe-claiming-app/src/widgets/ClaimingWidget.tsx
```diff
@@ -173,7 +180,14 @@ const ClaimingWidget = () => {
...
                 Delegated to
               </Typography>
-              <SelectedDelegate delegate={currentDelegate} />
+              <Link
+                href={claimingSafeAppUrl}
+                rel="noopener noreferrer"
+                target="_blank"
+                underline="none"
+              >
+                <SelectedDelegate delegate={currentDelegate} />
```
schmanu, line 189: Same here, it would be cool to have some feedback that this is clickable similar to how `ButtonBase` provides it.

schmanu: Alternatively we could wrap the entire widget in a clickable ButtonBase which opens the claiming app.

apps/safe-claiming-app/src/components/steps/Claim/index.tsx
DiogoSoaress: > I don't know whats with the buttons in the claiming app.

In this case, the button was hard to align with the TextField input given it always preserves the input field error height.

I am going to try your suggestion to see if we get to the same result 🙏

katspaugh: Why does the button have to be the same size? Can't it be just centered vertically?

apps/safe-claiming-app/src/widgets/ClaimingWidget.tsx
DiogoSoaress: > Alternatively we could wrap the entire widget in a clickable ButtonBase which opens the claiming app.

That almost defeats the purpose of having buttons in the widget. I will rather provide some visual feedback that the user understands which parts of the widget are clickable.

apps/safe-claiming-app/src/components/steps/Claim/index.tsx
DiogoSoaress: Hmm, after I remove the `size` property the button still not quite adjusts to the input height.
I'd go more pragmatic in this case and simply pass it the height the way it is. WDYT?

schmanu (approved):

francovenica:
The aligment of the button is fine
image

francovenica:
I'm not able to check if the widget in the dashboard is clickable cuz this ticket doesn't have a build created.

katspaugh (changes requested):
I've changed the config to point to this branch.
You can test it on https://safe-web-core.dev.5afe.dev/eth:0xA77DE01e157f9f57C7c4A326eeE9C4874D0598b6/home

The widget is linked now, but it opens in the iframe instead of a new tab.

apps/safe-claiming-app/src/widgets/ClaimingWidget.tsx
```diff
@@ -140,12 +148,11 @@ const ClaimingWidget = () => {
...
         <Title>Your voting power</Title>
-        <Box
-          display="flex"
-          alignItems="center"
-          justifyContent="center"
-          gap={1}
-          mb={1}
+        <StyledButtonLink
+          href={claimingSafeAppUrl}
+          rel="noopener noreferrer"
```
katspaugh, line 153: I would remove this, we will then see in the analytics how many people clicked on this.

DiogoSoaress: I will remove it then.
Also the CI failed as it does not allow me to pass the `<a>` props to the button even after passing the `href` .... I'm finding a solution

francovenica:
The widget works fine now, clicking both the balance and the current delegate takes you to the app in a new tab

PR #651, opened 2023-01-25
fix(tx-builder): disable tx simulation in Celo

dasanra:
## What it solves
Disable tx simulation for Celo

## How this PR fixes it
Adds Celo chain ID to disabled chains


mikhailxyz (approved):
🚀🚀🚀

mikhailxyz:
fyi this was reported in https://github.com/safe-global/safe-react-apps/issues/649

PR #657, opened 2023-02-01
feat(wallet-connect): update safe-apps-sdk to enable synchronous off-chain signatures

schmanu:
Depends on https://github.com/safe-global/safe-apps-sdk/pull/458
UI implementation: https://github.com/safe-global/web-core/pull/1630

### DIFF
--- a/apps/wallet-connect/src/hooks/useWalletConnectV1.tsx
+++ b/apps/wallet-connect/src/hooks/useWalletConnectV1.tsx
@@ -65,6 +65,7 @@ const useWalletConnectV1 = (
       })
 
       wcConnector.on('call_request', async (error, payload) => {
+        console.log('Call request:', payload)
         if (error) {
           throw error
         }
--- a/apps/wallet-connect/src/hooks/useWalletConnectV1.tsx
+++ b/apps/wallet-connect/src/hooks/useWalletConnectV1.tsx
@@ -65,7 +65,6 @@ const useWalletConnectV1 = (
       })
 
       wcConnector.on('call_request', async (error, payload) => {
-        console.log('Call request:', payload)
         if (error) {
           throw error
         }
--- a/apps/wallet-connect/src/hooks/useWalletConnectV2.tsx
+++ b/apps/wallet-connect/src/hooks/useWalletConnectV2.tsx
@@ -167,9 +167,11 @@ const useWalletConnectV2 = (
 
       // events
       web3wallet.on('session_proposal', async proposal => {
+        console.log('New session proposal', proposal)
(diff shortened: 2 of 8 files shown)


apps/wallet-connect/src/hooks/useWalletConnectV1.tsx
```diff
@@ -62,9 +62,13 @@ const useWalletConnectV1 = (
         trackEvent(NEW_SESSION_ACTION, WALLET_CONNECT_VERSION_1, wcConnector.peerMeta)
 
         setWcClientData(payload.params[0].peerMeta)
+
+        // HACK: Always enable off-chain msgs
+        web3Provider.send('safe_setSettings', [{ offChainSigning: true }])
```
katspaugh, line 67: @iamacook how can we make this work for OpenSea w/o this hack?
Basically, not just OpenSea, but any app that supports EIP-1271 will work out of the box with 1/X Safe.

iamacook: For future reference: after discussing this, we will look into defaulting OpenSea to off-chain for 1/n Safes.

apps/wallet-connect/src/hooks/useWalletConnectV2.tsx
```diff
@@ -168,8 +168,9 @@ const useWalletConnectV2 = (
       // events
       web3wallet.on('session_proposal', async proposal => {
         const { id, params } = proposal
-        const { requiredNamespaces } = params
+        const { requiredNamespaces, optionalNamespaces } = params
```
schmanu, line 171: We could remove this from this release as it's slightly out of scope.

But I realized that some dapps won't work without also taking the `optionalNamespaces` into account.

apps/wallet-connect/src/hooks/useWalletConnectV2.tsx
```diff
@@ -217,7 +218,7 @@ const useWalletConnectV2 = (
                 accounts: safeAccount || [
                   `${EVMBasedNamespaces}:${safe.chainId}:${safe.safeAddress}`,
                 ],
-                methods: EIP155Namespace.methods,
+                methods: [...EIP155Namespace.methods, ...(optionalEIP155Namespace?.methods || [])],
```
iamacook, line 221: If you use `concat`, you need not use a fallback (and it's also more performant).
```suggestion
                methods: EIP155Namespace.methods.concat(optionalEIP155Namespace?.methods),
```

iamacook (approved):

mmv08 (approved):
top content

PR #667, opened 2023-02-15
chore: new release

dasanra:
## chore
Update WalletConnect library version


JagoFigueroa:
Hey team, tested and verified for both v1 and v2. 

I ran into some issues when testing QR scanning on v2 via this [test dapp](https://lab.web3modal.com/v2Base) but after talking with the wallet connect team the issue is related to the QR modal switching colours on dark theme and that's why the scanning fails.

No action we will take regarding this as it should be fixed on the WC team end. In the future we can consider to add a library to support this "inversion" on dark mode but shouldn't be an issue.

Cheers!

JagoFigueroa (approved):

PR #671, opened 2023-02-21
feat(wallet-connect): EIP-1271 support

iamacook:
## What it solves

Adds support for EIP-1271 in the WalletConnect app by updating the related dependencies.

## How this PR fixes it

The `safe-apps-provider`, `safe-apps-sdk` and `safe-apps-react-sdk` that support EIP-1271 have been updated.

## How to test it

Prerequisites:
- Open the https://github.com/safe-global/web-core/pull/1162 deployment.
- Run the [EIP-1271 test DApp](https://github.com/5afe/eip-1271-dapp) locally.
  - `main` branch can be used to test WC v1
  - `wallet-connect-v2` branch can be used to test WC v2 (with the env vars specified in `env.sample`).

Method:
1. Open the WC app in the aforementioned deployment.
2. Connect to the aforementioned test DApp
3. Toggle off-chain signing and attempt to sign a message.
4. Observe the off-chain signing modal.
5. Sign the message.
6. Observe the signed message in the "Signatures" tab of the transaction list.

PR #674, opened 2023-02-23
chore(wallet-connect): new release

dasanra:
## Added
 - #671


JagoFigueroa:
Tested and verified that:

- EIP-1271 support works with the web PRs that have the implementation enabled.
- Nothing breaks on prod where the implementation is not yet live.
- WC V1 works as usual in prod.
- WC V2 works as usual in prod.

Maybe we can add #675 to the release as well and do a 2x1?

Cheers!

JagoFigueroa (approved):

PR #679, opened 2023-03-01
fix(e2e): Change Safe Apps URL

yagopv:
## What it solves
Changed safe-apps URL in the web breaks cypress tests

## How this PR fixes it
By updating the URL


JagoFigueroa (approved):

dasanra:
Let's use #678 . The Safe address is not set via query parameter too

PR #706, opened 2023-06-13
fix(wallet-connect): Removed workaround of accepting all accounts, methods & events in V2

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-react-apps/issues/705

## How this PR fixes it
1) Updating dependencies and retesting
2) Updating tests as we found type changes
3) Reviewing the [migration guide](https://docs.walletconnect.com/2.0/web/web3wallet/upgrade-guide) and looking for new API changes

## How to test it
Use https://react-app.walletconnect.com or any other WC v2 DApp

**Links**
[Changes in web3wallet monorepo since 1.2.0](https://github.com/WalletConnect/walletconnect-monorepo/compare/2.4.5...2.8.0)


Sparragus:
Hi everyone. I found an issue with the WC v2 integration. I documented it here: https://github.com/safe-global/safe-react-apps/issues/709

I'll be happy to discuss it and help as needed.

DaniSomoza (approved):

JagoFigueroa:
Looking good! I have tested it following this [doc](https://www.notion.so/safe-global/Wallets-WalletConnect-v2-Testing-Guide-EXTERNAL-5dd08804b2e141daa8bc3d5dd66ea1c5) with test data that was shared by the wc team and the results are the expected ones taking into account our implementation.

As we know, some connections won't work because of the new restrictions we added for the supported methods but that is the expected and secure behaviour to protect users on a smart contract wallet context.

Lets see how it evolves in the near future (I expect some reports from teams and users) but hopefully the wc team will have a better solution for our context soon.

PR #707, opened 2023-06-14
fix(tx-builder): Open a window instead download batch for Safari and Firefox

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-react-apps/issues/660

## How this PR fixes it
By using a blob and opening it in a new window (Firefox and Safari) or directly downloading (Chrome).
It is not the perfect solution but CSP policy is not allowing the download from the iframe. This is a [know bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1365502)

## How to test it
1) Test that the generated file is exactly the same and that can be exported/imported
2) Chrome should behave the same downloading the file
3) Firefox and Safari should open a new window showing the file and should allow to download it from there

### DIFF
--- a/apps/tx-builder/src/lib/storage.ts
+++ b/apps/tx-builder/src/lib/storage.ts
@@ -77,7 +77,7 @@ const downloadObjectAsJson = (batchFile: BatchFile) => {
   // https://bugzilla.mozilla.org/show_bug.cgi?id=1365502
   if (
     navigator.userAgent.includes('Firefox') ||
-    (navigator.userAgent.includes('Safari') && navigator.userAgent.includes('Chrome'))
+    (navigator.userAgent.includes('Safari') && !navigator.userAgent.includes('Chrome'))
   ) {
     return window.open(blobURL)
   }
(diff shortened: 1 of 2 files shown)


apps/tx-builder/src/lib/storage.ts
```diff
@@ -69,14 +69,24 @@ const getBatches = async () => {
...
-    encodeURIComponent(JSON.stringify(batchFile, stringifyReplacer))
+  const blobURL = URL.createObjectURL(
+    new Blob([JSON.stringify(batchFile, stringifyReplacer)], { type: 'application/json' }),
+  )
+
+  // If Firefox or Safari open a new window to download the file
+  // https://bugzilla.mozilla.org/show_bug.cgi?id=1365502
+  if (
+    navigator.userAgent.indexOf('Firefox') !== -1 ||
+    (navigator.userAgent.indexOf('Safari') !== -1 && navigator.userAgent.indexOf('Chrome') === -1)
```
dasanra, line 80: What about using `includes()` that already returns a `boolean` value? Also this is case sensitive

https://www.w3schools.com/jsref/jsref_includes.asp

PR #723, opened 2023-06-28
refactor(wallet-connect): Added walletconnect namespaces builder helper function

DaniSomoza:
## What it solves

Added [`buildApprovedNamespaces `](https://docs.walletconnect.com/2.0/web/web3wallet/wallet-usage#-namespaces-builder-util) helper function to parsing the required and optional namespaces.

## How this PR fixes it

```javascript
        const safeChain = `${EVMBasedNamespaces}:${safe.chainId}`
        const safeAccount = `${EVMBasedNamespaces}:${safe.chainId}:${safe.safeAddress}`
        const safeEvents = requiredEIP155Namespace.events // we accept all events like chainChanged & accountsChanged (even if they are not compatible with the Safe)

        const approvedSafeNamespaces = buildApprovedNamespaces({
          proposal: params,
          supportedNamespaces: {
            eip155: {
              chains: [safeChain], // only the Safe chain
              methods: compatibleSafeMethods, // only the Safe methods
              events: safeEvents,
              accounts: [safeAccount], // only the Safe account
            },
          },
        })
```

### DIFF
--- a/apps/wallet-connect/package.json
+++ b/apps/wallet-connect/package.json
@@ -4,11 +4,12 @@
   "private": true,
   "homepage": "./",
   "dependencies": {
-    "@safe-global/safe-apps-provider": "0.17.0",
     "@gnosis.pm/safe-react-components": "^0.9.7",
+    "@safe-global/safe-apps-provider": "0.17.0",
     "@safe-global/safe-gateway-typescript-sdk": "^3.7.3",
     "@walletconnect/client": "^1.8.0",
-    "@walletconnect/web3wallet": "^1.8.0",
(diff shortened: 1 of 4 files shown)


mmv08 (commented):
"chore" implies that no production code was changed https://stackoverflow.com/questions/26944762/when-to-use-chore-as-type-of-commit-message

DaniSomoza:
> "chore" implies that no production code was changed https://stackoverflow.com/questions/26944762/when-to-use-chore-as-type-of-commit-message

Then I should use `refactor` here right?

JagoFigueroa:
Looking good to me! 😉

PR #725, opened 2023-06-29
fix(wallet-connect): Wallet connect open dapp url is not updated

yagopv:
## What it solves
Resolves https://github.com/safe-global/safe-react-apps/issues/632, https://github.com/safe-global/safe-react-apps/issues/722


JagoFigueroa (approved):

JagoFigueroa:
Tested and verified!

PR #727, opened 2023-07-04
fix(wallet-connect): Enable dApps to use optional chains and avoid session rejection

DaniSomoza:
## What it solves

Now Safe chain is no mandatory in the `requiredNamespaces.chains` obj


JagoFigueroa (approved):

JagoFigueroa:
Tested and verified!

PR #752, opened 2023-08-22
Change logo color

lukasschor:
Replacing the image as we should not use the logo green-on-white and this image ranks high on Google image search so might cause others to use the green logo while we already removed it from the press kit as many used it as green-on-white.


dasanra:
Hey Lukas, will close this PR and add the necessary changes in this other one we are about to merge:
https://github.com/safe-global/safe-react-apps/pull/750/files

I updated to use SVG that is more efficient than PNG and also some old links I found in the README while doing that change.

Thank you for pointing this, also found this issue on other repos. Will change ASAP

PR #771, opened 2023-11-09
fix(ramp-network): add input to release action

schmanu:
## What it solves
The release action.yml was missing the input for the ramp network API key.

## How this PR fixes it
Adds it to the action.yml


usame-algan (approved):
Good catch

PR #772, opened 2023-11-09
chore: New release

schmanu:
# Fixes
fix(ramp-network): adds API key from .env variables (#768)


usame-algan (approved):
Looks good. Do we need to increase the version number or is it done automatically?

PR #801, opened 2024-04-18
feat(tx builder): use cgw for simulation configs

katspaugh:
## What it solves

Instead of using a hardcoded list of chains where simulation isn't supported, use CGW/config service to get the config of the current chain and check its features list.

## How this PR fixes it

I've replaced a hardcoded array of chain ids with an async call to CGW. I've added `useAsync` from the wallet repo for this.

Screenshot 2024-04-18 at 11 33 43


katspaugh:
Simulation is disabled on Aurora, so no button is shown:

Screenshot 2024-04-18 at 11 40 36

katspaugh:
Tested by @liliya-soroka ✅

PR #809, opened 2024-08-09
feat(tx-builder): autocomplete for contract method

katspaugh:
## What it solves
A user complained that a select offers a subpar UX when searching for a method name.

## How this PR fixes it
Replaces the Select component with an input with Autocomplete.

## How to test it
* Open Transaction Builder
* Enter a contract address with many methods
* Type the first letters of the method name

## Screenshots
Screenshot 2024-08-09 at 16 36 05


katspaugh:
Need to fix unit tests.

francovenica:
I tried to add the safe app as a custom one in dev with prod and stg cgw, in different networks but it always fails.
image

katspaugh:
Try opening some other app and replace the appUrl part.

francovenica:
Yeah, that worked. 
LGTM

PR #813, opened 2024-08-19
refactor(ramp): enable all chains

### LINKED ISSUE #811: Expand network support for Ramp Safe app
Currently the app is not visible on Optimism. There might be more chains that need to be added.

They support, among our chains:
Arbitrum
Avalanche
Celo
Ethereum
Linea
Optimism
Polygon
Polygon zkEVM
zkSync Era

Not supported:
Aurora
Scroll
XLayer

katspaugh:
## What it solves
Resolves #811

## How this PR fixes it
Removes the optional [`swapAsset`](https://docs.ramp.network/configuration#swapasset) parameter and with it the need to configure chains manually.

## How to test it
The app should still work as before.

### DIFF
--- a/apps/ramp-network/src/App.tsx
+++ b/apps/ramp-network/src/App.tsx
@@ -1,35 +1,13 @@
-import React, { useEffect, useState, useMemo } from 'react'
+import React, { useEffect, useState } from 'react'
 import { useSafeAppsSDK } from '@safe-global/safe-apps-react-sdk'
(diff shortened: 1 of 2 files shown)


katspaugh:
Closing because the `swapAsset` parameter is actually needed to limit what assets we show on each chain and exclude assets from other chains (including non-EVM assets like Bitcoin).

PR #821, opened 2024-09-25
feat(transaction-builder-theme): Remove safe-react-components from transaction builder

clovisdasilvaneto:
## What it solves
it solves the https://github.com/safe-global/safe-react-apps/issues/820 but this PR should only be merged when #822 and #823 gets merged on this branch.

## How this PR fixes it
It moves the icons from the safe-react-components repo to the transaction builder components folder in order to not use it from the achieved library anymore


clovisdasilvaneto:

usame-algan (approved):
Can we also remove the `@gnosis.pm/safe-react-components` dependency?

clovisdasilvaneto:
@usame-algan we can't because we still use some types of this package in the tx-builder.

PR #822, opened 2024-09-25
chore(tx-builder-theme): move safe react components to TX-builder components folder

### LINKED ISSUE #820: [Transction Builder] Update the theme to the latest from Safe{Wallet}
TxB is stuck with the old MUI theme from 2022, we should update it to use the same theme as used in the Safe{Wallet} main app.

Design: https://www.figma.com/design/tQyUQeAwxJkRfFmJCLU3wC/Safe-Apps-Redesign?node-id=2908-10267&t=M4BdtdYTY2lkbM5F-1

clovisdasilvaneto:
Current we're using an outdated theme in the transaction builder application. In order to have the right theme introduced on it, we need to first adjust the internal theme structure which we use in our components, and since the https://github.com/5afe/safe-react-components repo is achieved me and @katspaugh decided to move the components to the tx-builder/components folder and change the internal theme structure that the components are using.

## What it solves
Resolves #820

## How this PR fixes it
It moves the components from the safe-react-components repo to the tx-builder/components folder and introduce the right Safe theme into the application

## How to test it
- All unit tests from the tx-builder folder shuold pass. Currently we're not using it yet *A follow-up PR will be send after this one is merged*


katspaugh (approved):
Great work!

clovisdasilvaneto:
@francovenica I'll came up with another PR then 👍

PR #823, opened 2024-09-25
chore(tx-builder-theme): Remove safe-react components and theme from the tx-builder application

### LINKED ISSUE #820: [Transction Builder] Update the theme to the latest from Safe{Wallet}

clovisdasilvaneto:
Current we're using an outdated theme in the transaction builder application. In order to have the right theme introduced on it, we need to first adjust the internal theme structure which we use in our components, and since the https://github.com/5afe/safe-react-components repo is achieved me and @katspaugh decided to move the components to the tx-builder/components folder and change the internal theme structure that the components are using.

## What it solves
Resolves #820

## How this PR fixes it
It uses the moved safe-react-components in the tx-builder screens

## How to test it
- Open the safe wallet and add the tx-builder as a custom application, you should see the new theme colors as per figma design: https://www.figma.com/design/tQyUQeAwxJkRfFmJCLU3wC/Safe-Apps-Redesign?node-id=2908-10267&t=M4BdtdYTY2lkbM5F-1

Screenshot 2024-09-25 at 11 53 02


yarn.lock
```diff
@@ -2929,6 +2929,11 @@
   dependencies:
     semver "^7.3.7"
 
+"@safe-global/safe-gateway-typescript-sdk@^3.19.0":
+  version "3.22.2"
+  resolved "https://registry.yarnpkg.com/@safe-global/safe-gateway-typescript-sdk/-/safe-gateway-typescript-sdk-3.22.2.tgz#d4ff9972e58f9344fc95f8d41b2ec6517baa8e79"
+  integrity sha512-Y0yAxRaB98LFp2Dm+ACZqBSdAmI3FlpH/LjxOZ94g/ouuDJecSq0iR26XZ5QDuEL8Rf+L4jBJaoDC08CD0KkJw==
```
katspaugh, line 2935: Strange that this appeared. Is it due to `yarn install`?

katspaugh (approved):
Code looks good!

yarn.lock
clovisdasilvaneto: yes

francovenica:
Question. I see in the designs that the form should be on the full width of the view port. Is this only for small viewports?

This is how I currently see it
image

This is from the design:
image

This is the only difference I see, so if the width of the form is not part of the changes then the ticket is good to go

clovisdasilvaneto:
Hi @francovenica, yes the form width is not part of the change.

francovenica:
Ok, the ticket how it is can be passed to "ready to merge". Let me know if you can try that change you suggested here or if you will create a new PR for it. Thanks

PR #829, opened 2024-10-22
feat(tx-builder): Promote development to production

clovisdasilvaneto:
## What it solves
It promotes latest development branch code to main in order to promote it to production later.


.github/workflows/deployment.yml
```diff
@@ -17,7 +17,7 @@ env:
 jobs:
   pr:
     name: Pull Request
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
```
usame-algan, line 20: We should be able to use the latest version again since adding the python environment fix

usame-algan (approved):

